PR #3, opened 2020-06-22
Blobstore GET and POST, Setting up Servlet Structure

cassmarcussen:
In this PR, I make the following changes: 
- Add pom.xml dependencies for gson and Blobstore (appengine)
- Set up KeyframeImage.java class, the class for keyframe images
- Set up the safe search detection. However, I am unable to enable key creation for the Vision API as of now. Therefore, this code must be commented out for now, until key creation is enabled. This is a known issue among interns (in the known issues doc), so I will try to resolve this tomorrow.
- KeyframeImageUploadServlet and ServeKeyframeImage for Blobstore GET and POST method that work for both local and deployed versions of the webpage. Based on the tutorials for libraries and on my portfolio page code.
- keyFrameImageUpload.jsp: A site to perform POST requests. This is mostly for testing until we get all 3 features integrated, so I can hard-code POST requests to use and display images.
- Results html page with a div that is added to in the JavaScript with the images and effect.

Visuals: (Screenshot because SnipIt is currently not working)
Screenshot 2020-06-22 at 5 08 26 PM

<hr>

### Helper functions that generate html, from image-effect-results.js

htmlForEffect - creates the HTML for the label and meter for each SafeSearch effect category
[Visual](https://screenshot.googleplex.com/qiZs2rtWf4f)

createKeyframeImageTextInnerHTML - creates the HTML for the time metrics of the image (timestamp, start time, end time), and calls htmlForEffect to create the HTML for the meter for the effect
[Visual](https://screenshot.googleplex.com/pvopMEGzdpr)

setFlaggedImageSummaryContent - creates the HTML that displays how many keyframe images have been flagged for sensitive content by the Vision API’s SafeSearch detection method
[Visual](https://screenshot.googleplex.com/CKmGSyoiDe1)

setDisplayOfDots - sets the display of the dots that appear on the bottom of the keyframe image cards, to navigate between the keyframe images
[Visual](https://screenshot.googleplex.com/m3cUbcYmCqP)

createSingularKeyframeImageCard - calls helper functions to set up the HTML code for displaying the entire card for a keyframe image. Appends all the elements created to the div for the keyframe card.
[Visual](https://screenshot.googleplex.com/sBSUTFueE5r)

createKeyframeImageSlideshow - creates the slideshow of cards with keyframe images and their corresponding information and SafeSearch detected effect. It does so by iterating through the array of keyframe images returned from DataStore and calling createSingularKeyframeImageCard for each keyframe image to create a card in the slideshow for each flagged keyframe image.
I.e. controls the creation of all the keyframe image cards, where only the one is displayed and the rest are hidden 

showSlides - makes the slide at the current dot or index visible, and makes the rest hidden, i.e. the dots and arrows correspond to which keyframe image card is shown


cassmarcussen:
Updates:
- Use async/await and fetch in JavaScript to get the effect (hard-coded for now) corresponding to each image from the Java servlet. 
- Make the images show up in a carousel.

cassmarcussen:
Updates: 
- Use html meter tag to display the effects using bars instead of just text
- Store the effects as a HashMap<String, String> where the keys are the effects (adult, medical, spoofed, racy, violent), and the values are the likelihoods.

Notes: 
- Added in hard-coded values for the HashMap of effects. The key for the Cloud Vision API is not yet set up correctly, so all the values displayed are hard-coded. However, the process through which the variables are communicated between Java servlets and the front-end JS code is functional, not hard-coded.

Updated UI: 
ui_june24_capstone

cassmarcussen:
Updates: 
- Integrated Google Cloud Bucket with Blobstore. Now, images can be served on the Bucket (which is necessary for the Cloud Vision API), and their timestamp information can be served in DataStore.
- Now, keyframe images can be posted to a database, and can be retrieved from the database, displayed on the HTML page for Results, and passed to the Cloud Vision API. The results of the Cloud Vision API are retrieved and displayed visually on the HTML page using meters.
- Created a Delete function, so that at the start of the user flow the DataStore and Bucket can be emptied to avoid retrieving old data from past uses of the website.
- Updated the Results page's UI based on Rhiannan's design of the index page, for consistency across pages of Video Vigilance.

TODO:
- Test the POST method to post with Blobstore from another source file.
- The timestamp in DataStore and the images in Blobstore are sorted differently. The timestamps are sorted by String value (or, in order of earliest to latest time in the video). The images are sorted by alphabetical order for the Blobstore id's, which are arbitrarily generated. As of now, the timestamps and images do not line up, so this is a point of revision.

Updated UI: 
june30_results_ui

capstone/src/main/java/com/google/sps/servlets/DetectSafeSearchGcs.java
```diff
@@ -0,0 +1,71 @@
...
+    ImageSource imgSource = ImageSource.newBuilder().setGcsImageUri(gcsPath).build();
+    Image img = Image.newBuilder().setSource(imgSource).build();
+    Feature feat = Feature.newBuilder().setType(Type.SAFE_SEARCH_DETECTION).build();
+    AnnotateImageRequest request =
+        AnnotateImageRequest.newBuilder().addFeatures(feat).setImage(img).build();
+    requests.add(request);
+
+    // Initialize client that will be used to send requests. This client only needs to be created
+    // once, and can be reused for multiple requests. After completing all of your requests, call
+    // the "close" method on the client to safely clean up any remaining background resources.
```
rwackes, line 41: It says here that you want to call the "close" method, but I don't see the call in the code.

capstone/src/main/webapp/js/image-effect-results.js
```diff
@@ -0,0 +1,200 @@
...
+
+async function fetchBlobstoreKeyframeImages() {
+
+  fetch('/keyframe-image-upload', {method: 'GET'})
+    .then((response) => {
+        return response.text();
+    })
+    .then(async function (keyframeImages) {
+    
+        alert(keyframeImages);
```
rwackes, line 36: fix indentation

capstone/src/main/webapp/js/image-effect-results.js
```diff
@@ -0,0 +1,200 @@
...
+                var effectsAsNumbers = new Map()
+                effectsAsNumbers.set(effect.adult, getNumberOfEffectParameter(effect.adult));
+                effectsAsNumbers.set(effect.medical, getNumberOfEffectParameter(effect.medical));
+                effectsAsNumbers.set(effect.spoofed, getNumberOfEffectParameter(effect.spoofed));
+                effectsAsNumbers.set(effect.violence, getNumberOfEffectParameter(effect.violence));
+                effectsAsNumbers.set(effect.racy, getNumberOfEffectParameter(effect.racy));
+
+                // Don't display the image if it has no 4 or 5 (likely or very unlikely sensitive content), 
+                // i.e. only show the image if one of the effect parameters is 'likely' or 'very likely', and potentially 'possible'.
+                if(!Array.from(effectsAsNumbers.values()).includes(4) && !Array.from(effectsAsNumbers.values()).includes(5)) {
```
rwackes, line 91: Restructure if/else block so that if statement is not empty.

capstone/src/main/webapp/js/image-effect-results.js
```diff
@@ -0,0 +1,200 @@
...
+                    //continue;
+                }else {
+                    // Else, mark the image as flagged, i.e. increase the number of flagged images by one.
+                    numberOfFlaggedImages++;
+                }
+
+                var keyframeImageText = document.createElement("p");
+                keyframeImageText.innerHTML = '<h2>Information about the frame</h2>'
+                         + '<p>Timestamp of image: ' + timestamp + '</p>'
```
rwackes, line 100: fix indentation

capstone/src/main/webapp/js/image-effect-results.js
```diff
@@ -0,0 +1,200 @@
...
+                // Don't display the image if it has no 4 or 5 (likely or very unlikely sensitive content), 
+                // i.e. only show the image if one of the effect parameters is 'likely' or 'very likely', and potentially 'possible'.
+                if(!Array.from(effectsAsNumbers.values()).includes(4) && !Array.from(effectsAsNumbers.values()).includes(5)) {
+                    //continue;
+                }else {
+                    // Else, mark the image as flagged, i.e. increase the number of flagged images by one.
+                    numberOfFlaggedImages++;
+                }
+
+                var keyframeImageText = document.createElement("p");
```
rwackes, line 98: I believe that regardless of whether an image is flagged or not, this keyframeImageText paragraph element is displayed on the results page. This is confusing in the case when "You have no flagged images." Make sure this is only displayed on results webpage if image has a 4 or 5. But correct me if I am wrong.

capstone/src/main/java/com/google/sps/servlets/KeyframeImageDeleteServlet.java
```diff
@@ -0,0 +1,118 @@
...
+    deleteDataStoreInfo();
+    
+    deleteGoogleCloudBucketInfo();
+
+    // Redirect back to the HTML page.
+    response.sendRedirect("/index.html");
+
+  }
+
+  private void deleteDataStoreInfo() {
```
rwackes, line 60: Add comment for method.

capstone/src/main/java/com/google/sps/servlets/KeyframeImageDeleteServlet.java
```diff
@@ -0,0 +1,118 @@
...
+    PreparedQuery results = datastore.prepare(query);
+
+    for (Entity entity : results.asIterable()) {
+       datastore.delete(entity.getKey());
+    }
+
+  }
+
+  /* Reference: https://cloud.google.com/storage/docs/deleting-objects#storage-delete-object-java */
+  private void deleteGoogleCloudBucketInfo() {
```
rwackes, line 75: Add comment for method.

capstone/src/main/java/com/google/sps/servlets/KeyframeImageDeleteServlet.java
```diff
@@ -0,0 +1,118 @@
...
+
+        storage.delete(bucketName, objectName);
+
+        System.out.println("Object " + objectName + " was deleted from " + bucketName);
+    }
+
+
+  }
+
+  public static ArrayList<String> listObjects() {
```
rwackes, line 97: Add comment for method.

capstone/src/main/java/com/google/sps/data/KeyframeImage.java
```diff
@@ -0,0 +1,50 @@
+package com.google.sps.servlets;
+
```
sharond106, line 2: I think it will be helpful to add a comment briefly describing what the KeyframImage class does

capstone/src/main/java/com/google/sps/servlets/DetectSafeSearchGcs.java
```diff
@@ -0,0 +1,71 @@
...
+import com.google.cloud.vision.v1.ImageAnnotatorClient;
+import com.google.cloud.vision.v1.ImageSource;
+import com.google.cloud.vision.v1.SafeSearchAnnotation;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.HashMap;
+
+public class DetectSafeSearchGcs {
+
```
sharond106, line 18: Same here for adding a comment

capstone/src/main/java/com/google/sps/servlets/ImageEffectServlet.java
```diff
@@ -0,0 +1,29 @@
...
+import com.google.gson.Gson;
+import java.io.IOException;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import java.util.HashMap;
+
+/** Servlet that returns some example content. TODO: modify this file to handle comments data */
+@WebServlet("/keyframe-effect-servlet")
```
sharond106, line 12: This comment just needs updated

capstone/src/main/java/com/google/sps/servlets/KeyframeImageUploadServlet.java
```diff
@@ -0,0 +1,128 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.ArrayList;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+@WebServlet("/keyframe-image-upload")
+public class KeyframeImageUploadServlet extends HttpServlet {
```
sharond106, line 32: add comment

capstone/src/main/java/com/google/sps/servlets/KeyframeImageUploadServlet.java
```diff
@@ -0,0 +1,128 @@
...
+    // Get the startTime
+    String startTime = request.getParameter("startTime");
+    // Get the endTime
+    String endTime = request.getParameter("endTime");
+
+    //Check for null, do not do post request if null url
+    if (imageUrl == null || imageUrl.contains("undefined")) {
+        response.sendRedirect("js/keyframeImageUpload.jsp");
+        return;
+    }
```
sharond106, line 91: We should also decide what we want to do in case the time stamps are null. Maybe here you can add a check for the ordering of the time stamps for now (startTime <= timestamp <= endTime)

capstone/src/main/java/com/google/sps/servlets/KeyframeImageUploadServlet.java
```diff
@@ -0,0 +1,128 @@
...
+
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    datastore.put(entity);
+
+    response.sendRedirect("/results.html");
+  
+  }
+
+  /** Returns a URL that points to the uploaded file, or null if the user didn't upload a file. */
+  private String getUploadedFileUrl(HttpServletRequest request, String formInputElementName) {
```
sharond106, line 108: I think you've used "//" for commenting functions before. Keep it consistent?

capstone/src/main/java/com/google/sps/servlets/ServeKeyFrameImage.java
```diff
@@ -0,0 +1,25 @@
...
+import com.google.appengine.api.blobstore.BlobstoreService;
+import com.google.appengine.api.blobstore.BlobstoreServiceFactory;
+import java.io.IOException;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+@WebServlet("/serve")
+public class ServeKeyFrameImage extends HttpServlet {
```
sharond106, line 11: add comment for servlet

capstone/src/main/java/com/google/sps/servlets/ServeKeyFrameImage.java
```diff
@@ -0,0 +1,25 @@
...
+    @Override
+    public void doGet(HttpServletRequest request, HttpServletResponse response)
+        throws IOException {
+            BlobKey blobKey = blobstoreService.createGsBlobKey(
+                "/gs/keyframe-images-for-effect/" + fileNum + "-keyframe-image");
+            fileNum++;
+            blobstoreService.serve(blobKey, response);
+
+            
+        }
```
sharond106, line 24: Where is this servlet being used? I can't find where it's being called. I was also wondering how the fileNum works- will it ever decrease/ get reset?

capstone/src/main/webapp/css/results.css
```diff
@@ -0,0 +1,138 @@
+/*body {
+    font-family: 'Roboto', sans-serif;
+}*/
+
+img {
+    width: 750px;
+}
```
sharond106, line 7: Not sure how much CSS code style we need to adhere to, but your tabs are sometimes 2 spaces and sometimes 4 spaces. The style attributes are also supposed to be in alphabetical order, but maybe save that for later!

capstone/src/main/webapp/js/image-effect-results.js
```diff
@@ -0,0 +1,200 @@
...
+    } else if (effectParameter == "POSSIBLE") {
+        numberOfEffect = 3;
+    } else if (effectParameter == "LIKELY") {
+        numberOfEffect = 4;
+    } else if (effectParameter == "VERY_LIKELY") {
+        numberOfEffect = 5;
+    }
+
+    return numberOfEffect;
+}
```
sharond106, line 26: This is basically just creating an enum for the effect right? It might be more concise to declare a real enum like: https://stackoverflow.com/questions/287903/what-is-the-preferred-syntax-for-defining-enums-in-javascript

capstone/src/main/webapp/js/image-effect-results.js
```diff
@@ -0,0 +1,200 @@
...
+                        + '<p><label for="medical">Medical: ' + effect.medical + '</label> \
+                            <meter id="medical" value="' + effectsAsNumbers.get(effect.medical) + '"  min="0" low="3" high="4" optimum="5" max="5"></meter></p>'
+                        + '<p><label for="spoofed">Spoofed: ' + effect.spoofed + '</label> \
+                            <meter id="spoofed" value="' + effectsAsNumbers.get(effect.spoofed) + '"  min="0" low="3" high="4" optimum="5"  max="5"></meter></p>'
+                        + '<p><label for="violence">Violence: ' + effect.violence + '</label> \
+                            <meter id="violence" value="' + effectsAsNumbers.get(effect.violence) + '"   min="0" low="3" high="4" optimum="5"  max="5"></meter></p>'
+                        + '<p><label for="racy">Racy: ' + effect.racy + '</label> \
+                            <meter id="racy" value="' + effectsAsNumbers.get(effect.racy) + '"   min="0" low="3" high="4" optimum="5"  max="5"></meter></p>'
+                        + '<p>Likeliness values are Unknown, Very Unlikely, Unlikely, Possible, Likely, and Very Likely</p>';
```
sharond106, line 115: Could a helper function be used for each of these labels? It seems like you could pass in a bunch of parameters to a function that writes the HTML code instead of hardcoding it multiple times

capstone/src/main/webapp/js/image-effect-results.js
```diff
@@ -0,0 +1,200 @@
...
+}
+
+// Thumbnail image controls
+function currentSlide(n) {
+  showSlides(slideIndex = n);
+}
+
+function showSlides(n) {
+  var i;
+  var slides = document.getElementsByClassName("mySlides");
```
sharond106, line 187: Your tabs here are different from above

sharond106:
A general comment I had is to just write more comments in your code. Some classes, functions and big loops did not have one. But good job- it looks super cool!

capstone/src/main/webapp/js/image-effect-results.js
cassmarcussen: The 'if' statement is only empty for testing, so that all keyframe images are displayed instead of those just flagged for negative effect. I'll leave a comment indicating that this is just for testing temporarily.

capstone/src/main/webapp/js/image-effect-results.js
cassmarcussen: You are correct! For now, this is just to test more images, regardless of if they are flagged. Once I uncomment the 'continue' in an 'if' statement in fetchBlobstoreKeyframeImages, only flagged images will be displayed, which I have tested out.

capstone/src/main/java/com/google/sps/servlets/ServeKeyFrameImage.java
cassmarcussen: Thanks for pointing this out! I think this servlet should be deleted, so I'll make this change.

capstone/src/main/java/com/google/sps/servlets/KeyframeImageUploadServlet.java
cassmarcussen: [Note for future me:] I'll use the "/*" comment at the top of functions (i.e. to describe functions), and "//" for comments within functions.

capstone/src/main/java/com/google/sps/data/KeyframeImage.java
```diff
@@ -0,0 +1,54 @@
...
+Google Cloud Bucket URL of the image, the timestamp of the image, and the start and end time of the shot that the image is from. 
+This class implements Comparable, so that it can have a compareTo method which can be used for sorting the KeyframeImages by timestamp.
+*/
+public class KeyframeImage implements Comparable<KeyframeImage>{
+
+    private String url;
+    //will these be Strings?
+    private String timestamp;
+    private String startTime;
+    private String endTime;
```
sharond106, line 13: My times are all numbers right now. I'm not sure if they are automatically converted to strings when we send them to the POST method but we can figure that out later!

capstone/src/main/webapp/js/image-effect-results.js
```diff
@@ -0,0 +1,210 @@
...
+        keyframeImageDiv.classList.add("mySlides");
+        keyframeImageDiv.classList.add("keyframe-card-fade");
+
+        var thisImage = arrayOfKeyframeImages[i];
+
+        var keyframeImage = document.createElement("img");
+        keyframeImage.src = thisImage.url.replace("gs://", "https://storage.cloud.google.com/");
+
+        if(keyframeImage.src != null && keyframeImage.src.indexOf("undefined") == -1){
+
```
sharond106, line 78: space after before and after (...)

capstone/src/main/webapp/js/image-effect-results.js
```diff
@@ -0,0 +1,210 @@
...
+          effectsAsNumbers.set(effect.violence, getNumberOfEffectParameter(effect.violence));
+          effectsAsNumbers.set(effect.racy, getNumberOfEffectParameter(effect.racy));
+
+          // Don't display the image if it has no 4 or 5 (likely or very unlikely sensitive content), 
+          // i.e. only show the image if one of the effect parameters is 'likely' or 'very likely', and potentially 'possible'.
+          if(!Array.from(effectsAsNumbers.values()).includes(4) && !Array.from(effectsAsNumbers.values()).includes(5)) {
+            // continue is commented out temporarily for testing, so that all keyframe images are displayed instead of just those flagged for negative effect
+            //continue;
+          }else {
+            // Else, mark the image as flagged, i.e. increase the number of flagged images by one.
```
sharond106, line 102: space after 'if' and before 'else'

sharond106:
There were some old comments that weren't resolved- I'm not sure if you wanted to change those or not, but everything else looks good to me!

capstone/src/main/java/com/google/sps/data/KeyframeImage.java
cassmarcussen: What format are the numbers for time in? I think sending them as Strings would be ideal, if possible, since all we are doing with the times for now is displaying them on the page to the user in a nice format.

However, if we want to make the times a hyperlink to that time in the video (which I've been thinking about), maybe sending them as numbers is best.

Anyways, we can decide this later but just some ideas!

capstone/src/main/java/com/google/sps/data/KeyframeImage.java
```diff
@@ -0,0 +1,54 @@
+package com.google.sps.servlets;
+
+/* The KeyframeImage class holds information about each keyframe image extracted from the video advertisement, including the 
+Google Cloud Bucket URL of the image, the timestamp of the image, and the start and end time of the shot that the image is from. 
+This class implements Comparable, so that it can have a compareTo method which can be used for sorting the KeyframeImages by timestamp.
+*/
+public class KeyframeImage implements Comparable<KeyframeImage>{
+
+    private String url;
```
vkkpng-g, line 9: To make this more clear, lets rename to cloudBucketUrl. 

You have great comments so I understood it, but in code people wouldn't be reading the KeyFrameImage class definition to figure out what url really it. Therefore having a clearer names makes it easier for other people.

capstone/src/main/java/com/google/sps/data/KeyframeImage.java
vkkpng-g: I haven't looked at Sharon PR yet, but I think in the user perspective seeing a string time stamp of like 12:03 is more preferable.

capstone/src/main/java/com/google/sps/data/KeyframeImage.java
```diff
@@ -0,0 +1,54 @@
...
+    //will these be Strings?
+    private String timestamp;
+    private String startTime;
+    private String endTime;
+    private String effect;
```
vkkpng-g, line 14: In your class definition, I don't believe you discussed what the effect is.

Do you think a better naming like videoEffect/UserEffect would be better? Something along those lines what kind of effect this class is encapsulating.

capstone/src/main/java/com/google/sps/data/KeyframeImage.java
```diff
@@ -0,0 +1,54 @@
...
+        return effect;
+    }
+
+    public void setEffect(String newEffect) {
+        effect = newEffect;
+    }
+
+    // compareTo for KeyframeImage, where images with earlier timeframes are considered less than images with later timeframes
+    @Override
+    public int compareTo(KeyframeImage o){
```
vkkpng-g, line 50: When is this function used and why are we considering images with earlier timeframes less. I assume its because ur doing some sort of ordering, but add some clarifications why we are comparing a certain way.

capstone/src/main/java/com/google/sps/data/KeyframeImage.java
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    public void setEffect(String newEffect) {
+        effect = newEffect;
+    }
+
+    // compareTo for KeyframeImage, where images with earlier timeframes are considered less than images with later timeframes
+    @Override
+    public int compareTo(KeyframeImage o){ 
+        return this.timestamp.compareTo(o.getTimestamp());
```
vkkpng-g, line 51: Do we have to include "this" for this method to work correctly. The reason why I'm asking is because for all the function above you don't include it. It's good to have consistency.

capstone/src/main/java/com/google/sps/servlets/DetectSafeSearchGcs.java
```diff
@@ -0,0 +1,79 @@
...
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.HashMap;
+
+/* The DetectSafeSearchGcs class handles the user flow component of extracting the effect from a keyframe image. 
+It calls the Cloud Vision API's SafeSearch method to get the effect of the image, in terms of the parameters 
+of 'adult', 'medical', 'spoofed', 'violence', and 'racy'.
+*/
+public class DetectSafeSearchGcs {
```
vkkpng-g, line 21: Can we add comment what GCS stands for?

capstone/src/main/java/com/google/sps/servlets/DetectSafeSearchGcs.java
```diff
@@ -0,0 +1,79 @@
...
+
+  /* detectSafeSearchGcs with a KeyframeImage parameter has the same functionality as the function below with a String parameter. 
+  See the function below for a description. This function calls the one below, passing in the image's url as a parameter. */
+  public static HashMap<String, String> detectSafeSearchGcs(KeyframeImage image) throws IOException {
```
vkkpng-g, line 25: Is there a reason we want to do this?  I feel like this method is just parsing which is fine moving to the method below (combine)

capstone/src/main/java/com/google/sps/servlets/DetectSafeSearchGcs.java
```diff
@@ -0,0 +1,79 @@
...
+    Image img = Image.newBuilder().setSource(imgSource).build();
+    Feature feat = Feature.newBuilder().setType(Type.SAFE_SEARCH_DETECTION).build();
+    AnnotateImageRequest request =
+        AnnotateImageRequest.newBuilder().addFeatures(feat).setImage(img).build();
+    requests.add(request);
+
+    // Initialize client that will be used to send requests. This client only needs to be created
+    // once, and can be reused for multiple requests. After completing all of your requests, call
+    // the "close" method on the client to safely clean up any remaining background resources.
+    try (ImageAnnotatorClient client = ImageAnnotatorClient.create()) {
```
vkkpng-g, line 47: Do we not expect there to be errors?

capstone/src/main/java/com/google/sps/servlets/DetectSafeSearchGcs.java
```diff
@@ -0,0 +1,79 @@
...
+      BatchAnnotateImagesResponse response = client.batchAnnotateImages(requests);
+      List<AnnotateImageResponse> responses = response.getResponsesList();
+
+      for (AnnotateImageResponse res : responses) {
+        if (res.hasError()) {
+          System.out.format("Error: %s%n", res.getError().getMessage());
+          safeSearchResults.put("adult", res.getError().getMessage());
```
vkkpng-g, line 54: Why not just return an empty array?

capstone/src/main/java/com/google/sps/servlets/KeyframeImageDeleteServlet.java
```diff
@@ -0,0 +1,126 @@
...
+// limitations under the License.
+
+package com.google.sps.servlets;
+
+import com.google.api.gax.paging.Page;
+import com.google.cloud.storage.Blob;
+import com.google.cloud.storage.Bucket;
+import com.google.cloud.storage.Storage;
+import com.google.cloud.storage.StorageOptions;
+import com.google.appengine.tools.cloudstorage.GcsFilename;
```
vkkpng-g, line 22: move above (alphabetized)

capstone/src/main/java/com/google/sps/servlets/KeyframeImageDeleteServlet.java
```diff
@@ -0,0 +1,126 @@
...
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import java.util.ArrayList;
+import java.util.List;
+
+
+/** Servlet responsible for deleting keyframe images. This will be called at the beginning of the user flow, to ensure 
+that the DataStore list and the Google Cloud Bucket are emptied of any previous data.*/
+@WebServlet("/keyframe-image-delete")
```
vkkpng-g, line 45: This may be too complex for the first version, but what happens when 3 people go on ur website at the same time and all upload a video. 

Is this design assuming that only 1 person will be using ur application one at a time.

capstone/src/main/java/com/google/sps/servlets/KeyframeImageDeleteServlet.java
```diff
@@ -0,0 +1,126 @@
...
+
+    // Redirect back to the HTML page.
+    response.sendRedirect("/index.html");
+
+  }
+
+  /* Deletes all objects from the DataStore list associated with the keyframe images for the video.
+  */
+  private void deleteDataStoreInfo() {
+    String queryType = "KeyframeImages_Video";
```
vkkpng-g, line 66: Can we define what queryType is. Is this hardcoded for now before we integrate user input?

capstone/src/main/java/com/google/sps/servlets/KeyframeImageDeleteServlet.java
```diff
@@ -0,0 +1,126 @@
...
+    Storage storage = StorageOptions.newBuilder().setProjectId(projectId).build().getService();
+
+    List<String> listOfObjectNames = listObjects();
+
+    // objectName is the ID of your GCS object
+    for (String objectName : listOfObjectNames) {
+
+        storage.delete(bucketName, objectName);
+
+        System.out.println("Object " + objectName + " was deleted from " + bucketName);
```
vkkpng-g, line 98: Remove please.

capstone/src/main/java/com/google/sps/servlets/KeyframeImageDeleteServlet.java
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+  }
+
+  /* Lists the objects in the Google Cloud Bucket associated with the keyframe images, 
+  so that they can be deleted from the Bukcet in deleteGoogleCloudBucketInfo.
+  */
+  public static ArrayList<String> listObjects() {
```
vkkpng-g, line 106: So the items aren't keyframe images, but is there a better name?

capstone/src/main/java/com/google/sps/servlets/KeyframeImageUploadServlet.java
```diff
@@ -0,0 +1,136 @@
...
+ to make it a viable Google Cloud Storage Bucket url, which is necessary for using the Vision API.
+ */
+ @Override
+ public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+
+    Gson gson = new Gson();
+
+    List<KeyframeImage> keyframeImagesFromVideo = new ArrayList<>();
+
+    Query query = new Query("KeyframeImages_Video");
```
vkkpng-g, line 49: Is KeyframeImages_Video a constant or will it be replaced with user input. If it is constant, please define it above as a constant.

capstone/src/main/java/com/google/sps/servlets/KeyframeImageUploadServlet.java
```diff
@@ -0,0 +1,136 @@
...
+    query.addSort("timestamp",
+                     Query.SortDirection.ASCENDING);
```
vkkpng-g, line 51: Add comment why we are sorting in ASCENDING

capstone/src/main/java/com/google/sps/servlets/KeyframeImageUploadServlet.java
```diff
@@ -0,0 +1,136 @@
...
+
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    PreparedQuery results = datastore.prepare(query);
+
+    for (Entity entity : results.asIterable()) {
+
+      String urlForGCS = (String) entity.getProperty("url");
+
+      String url = "gs:/" + urlForGCS;
```
vkkpng-g, line 60: add gs as a constant

Maybe like DefaultPathForBlahBlah

capstone/src/main/java/com/google/sps/servlets/KeyframeImageUploadServlet.java
```diff
@@ -0,0 +1,136 @@
...
+
+      String timestamp = (String) entity.getProperty("timestamp");
+      long id = entity.getKey().getId();
+      String startTime = (String) entity.getProperty("startTime");
+      String endTime = (String) entity.getProperty("endTime");
+
+      KeyframeImage img = new KeyframeImage(url, timestamp, startTime, endTime);
+
+      keyframeImagesFromVideo.add(img);
+
```
vkkpng-g, line 70: For all the timestamp, id, starttime, endtime are we sure that the server will def return a valid value. Will any be missing?

capstone/src/main/java/com/google/sps/servlets/KeyframeImageUploadServlet.java
```diff
@@ -0,0 +1,136 @@
...
+    datastore.put(entity);
+
+    response.sendRedirect("/results.html");
+  
+  }
+
+  /** Returns a Google Cloud Storage Bucket URL that points to the uploaded file, or null if the user didn't upload a file. */
+  private String getUploadedFileUrl(HttpServletRequest request, String formInputElementName) {
+    BlobstoreService blobstoreService = BlobstoreServiceFactory.getBlobstoreService();
+    Map<String, List<BlobKey>> blobs = blobstoreService.getUploads(request);
```
vkkpng-g, line 118: Add comment what is String and List<BlobKey>

capstone/src/main/webapp/js/image-effect-results.js
```diff
@@ -0,0 +1,210 @@
...
+/* getNumberOfEffectParameter returns a number corresponding to the effect likelihood of 
+the keyframe image. This is used for the html meter which visually displays the likelihood 
+of each SafeSearch parameter on the page. The numbers returned are used to fill in the meter by 
+a certain amount (the number returned divided by 5).
+*/
+function getNumberOfEffectParameter(effectParameter) {
+
+  var numberOfEffect = 0;
+    
+  if (effectParameter == "UNKNOWN") {
```
vkkpng-g, line 16: Use Switch case

capstone/src/main/webapp/js/image-effect-results.js
```diff
@@ -0,0 +1,210 @@
...
+
+  return numberOfEffect;
+}
+
+/* fetchBobstoreKeyframeImages calls the GET method of the KeyframeImageUploadServlet to get the 
+keyframe images from DataStore and the Google Cloud Bucket. It then gets the image's effect using 
+the Google Cloud Vision API (called from Java), and displays keyframe images that are flagged for 
+possible, likely, or very likely sensitive content.
+*/
+async function fetchBlobstoreKeyframeImages() {
```
vkkpng-g, line 38: This method is very long, are you able to separate out the contents?

capstone/src/main/webapp/js/image-effect-results.js
```diff
@@ -0,0 +1,210 @@
...
+        var keyframeImageDiv = document.createElement("div");
+        keyframeImageDiv.classList.add("mySlides");
+        keyframeImageDiv.classList.add("keyframe-card-fade");
+
+        var thisImage = arrayOfKeyframeImages[i];
+
+        var keyframeImage = document.createElement("img");
+        keyframeImage.src = thisImage.url.replace("gs://", "https://storage.cloud.google.com/");
+
+        if (keyframeImage.src != null && keyframeImage.src.indexOf("undefined") == -1) {
```
vkkpng-g, line 77: Add comment what this condition is

capstone/src/main/webapp/js/image-effect-results.js
```diff
@@ -0,0 +1,210 @@
...
+          if (!Array.from(effectsAsNumbers.values()).includes(4) && !Array.from(effectsAsNumbers.values()).includes(5)) {
+            // continue is commented out temporarily for testing, so that all keyframe images are displayed instead of just those flagged for negative effect
+            //continue;
+          } else {
+            // Else, mark the image as flagged, i.e. increase the number of flagged images by one.
+            numberOfFlaggedImages++;
+          }
+
+          var keyframeImageText = document.createElement("p");
+          keyframeImageText.innerHTML = '<h2>Information about the frame</h2>'
```
vkkpng-g, line 107: Can we create a new method where you pass in "effect"

This method will return the html

capstone/src/main/webapp/js/image-effect-results.js
```diff
@@ -0,0 +1,210 @@
...
+              keyframeImageDiv.style.display = "block";
+              document.getElementById("dots").innerHTML += '<span class="dot active" onclick="currentSlide(' + (i + 1) + ')"></span>';
+            } else {
+              document.getElementById("dots").innerHTML += '<span class="dot" onclick="currentSlide(' + (i + 1) + ')"></span>';
+            }
+        }
+
+      }
+
+      if (numberOfFlaggedImages > 0) {
```
vkkpng-g, line 142: you can just have a variable

var flaggedImageText = blah blah
var notFlaggedImageText =  blah blah 

var comment = numberOfFlaggedImages > 0 ?  flaggedImageText : notFlaggedImage
document.getElementById("results-overview").innerHTML = comment

This makes it clearer

capstone/src/main/java/com/google/sps/servlets/KeyframeImageDeleteServlet.java
wdi-gg, line 106: I'm not sure if it's necessary to create a separate function? If we combine the methods, we won't need to repeat the first few lines, and we can loop the blobs and delete them directly, rather than loop blobs to get their names and loop again to delete.

capstone/src/main/java/com/google/sps/servlets/KeyframeImageDeleteServlet.java
wdi-gg: Good point.

capstone/src/main/java/com/google/sps/servlets/KeyframeImageUploadServlet.java
```diff
@@ -0,0 +1,136 @@
...
+ /* 
+ The GET method is used to get each entity from the DataStore database. The url of the entity returned is given a "gs:/" at the beginning 
+ to make it a viable Google Cloud Storage Bucket url, which is necessary for using the Vision API.
+ */
+ @Override
+ public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+
+    Gson gson = new Gson();
+
+    List<KeyframeImage> keyframeImagesFromVideo = new ArrayList<>();
```
wdi-gg, line 47: Declared when needed?

capstone/src/main/java/com/google/sps/servlets/KeyframeImageUploadServlet.java
```diff
@@ -0,0 +1,136 @@
...
+    PreparedQuery results = datastore.prepare(query);
+
+    for (Entity entity : results.asIterable()) {
+
+      String urlForGCS = (String) entity.getProperty("url");
+
+      String url = "gs:/" + urlForGCS;
+
+      String timestamp = (String) entity.getProperty("timestamp");
+      long id = entity.getKey().getId();
```
wdi-gg, line 63: Where do we use this id?

capstone/src/main/java/com/google/sps/data/KeyframeImage.java
wdi-gg: Making the times a hyperlink to that time in the video sounds like a brilliant idea! Either way, I think it's better to save time as a number for the nature of the variable, and consistency with Sharon's implementation. Besides, we should be able to easily convert number into nice looking string timestamp in js code.

capstone/src/main/java/com/google/sps/data/KeyframeImage.java
cassmarcussen: I was initially using this function to sort Keyframe Images after retrieving them from DataStore. However, I changed my implementation and am no longer using this.

I thought I should maybe keep this function in case it needs to be used in the future, for extensibility. However, it would be okay to delete, since it's not being used. Would you recommend deleting this function?

capstone/src/main/java/com/google/sps/data/KeyframeImage.java
cassmarcussen: No, it's not needed, so I'll remove the "this".

capstone/src/main/java/com/google/sps/servlets/DetectSafeSearchGcs.java
cassmarcussen: I think I should delete this method. It is not used and is just a method which parses and calls detectSafeSearchGcs(String gcsPath).

capstone/src/main/java/com/google/sps/servlets/KeyframeImageDeleteServlet.java
cassmarcussen: Good point. I have been thinking about addressing this by assigning a unique bucket name for each user of the application/each video upload. I'll think more about how to address this as I work on improving and optimizing my feature.

capstone/src/main/java/com/google/sps/servlets/KeyframeImageDeleteServlet.java
cassmarcussen: The queryType defines the DataStore list that we should reference to access the keyframe images stored. This will be hardcoded always, unless we assign a unique bucket name for each user to account for multiple users using the application at the same itme.

capstone/src/main/java/com/google/sps/servlets/KeyframeImageDeleteServlet.java
cassmarcussen: The items are the names of the keyframe images as they are stored in the Google Cloud Storage Bucket. I will rename this to listNamesOfKeyframeImages().

capstone/src/main/java/com/google/sps/servlets/KeyframeImageUploadServlet.java
cassmarcussen: It is a constant, so I'll put 'final' in front of it.

capstone/src/main/java/com/google/sps/servlets/KeyframeImageDeleteServlet.java
cassmarcussen: Good point. Combining the functions also saves memory since we don't need to create an ArrayList<String>

capstone/src/main/java/com/google/sps/servlets/KeyframeImageUploadServlet.java
cassmarcussen: It's not used - I'll delete it.

capstone/src/main/java/com/google/sps/data/KeyframeImage.java
```diff
@@ -0,0 +1,48 @@
...
+
+    private String cloudBucketUrl;
+    // The following three int variables are times in number of seconds.
+    private int timestamp;
+    private int startTime;
+    private int endTime;
+    private String safeSearchEffect;
+
+    public KeyframeImage(String myUrl, int myTimestamp, int myStartTime, int myEndTime) {
+        cloudBucketUrl = myUrl;
```
vkkpng-g, line 18: nit why do we name the parameter with "myBlah"

capstone/src/main/java/com/google/sps/servlets/DetectSafeSearchGcs.java
```diff
@@ -0,0 +1,75 @@
...
+*/
+public class DetectSafeSearchGcs {
+
+  /* Detects whether the specified image on Google Cloud Storage has features you would want to moderate. */
+  public static HashMap<String, String> detectSafeSearchGcs(String gcsPath) throws IOException {
+
+   // String gcsUrl = "gs://keyframe-images-for-effect/nyc.jpg";
+    ImageSource imgSource = ImageSource.newBuilder().setGcsImageUri(gcsPath).build();
+    Image img = Image.newBuilder().setSource(imgSource).build();
+    Feature feat = Feature.newBuilder().setType(Type.SAFE_SEARCH_DETECTION).build();
```
vkkpng-g, line 32: nit: lets name this feature

capstone/src/main/java/com/google/sps/servlets/KeyframeImageUploadServlet.java
```diff
@@ -0,0 +1,143 @@
...
+ public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+
+    // queryType defines the DataStore list that we should reference to access the keyframe images stored
+    final String queryType = "KeyframeImages_Video";
+    Query query = new Query(queryType);
+
+    // We sort in ASCENDING so that the timestamps are sorted from earliest to latest. This ensures that keyframe
+    // images which are earlier in the video ad are shown in the slideshow of images on the Results page earlier.
+    query.addSort("timestamp",
+                     Query.SortDirection.ASCENDING);
```
vkkpng-g, line 52: move the Query.SortDirection back up?

capstone/src/main/java/com/google/sps/data/KeyframeImage.java
cassmarcussen: Fixed - renamed to "inputBlah"

capstone/src/main/java/com/google/sps/servlets/DetectSafeSearchGcs.java
cassmarcussen: Fixed - named safeSearchDetectionFeature

capstone/src/main/java/com/google/sps/servlets/KeyframeImageUploadServlet.java
cassmarcussen: Fixed

capstone/src/main/java/com/google/sps/servlets/DetectSafeSearchGcs.java
```diff
@@ -0,0 +1,79 @@
...
+    // Initialize client that will be used to send requests. This client only needs to be created
+    // once, and can be reused for multiple requests. After completing all of your requests, call
+    // the "close" method on the client to safely clean up any remaining background resources.
+    try (ImageAnnotatorClient client = ImageAnnotatorClient.create()) {
+      BatchAnnotateImagesResponse response = client.batchAnnotateImages(requests);
+      List<AnnotateImageResponse> responses = response.getResponsesList();
+
+      for (AnnotateImageResponse res : responses) {
+        if (res.hasError()) {
+          System.out.format("Error: %s%n", res.getError().getMessage());
```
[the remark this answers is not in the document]
wdi-gg: Let's remove the print if we are not using it?

capstone/src/main/webapp/js/image-effect-results.js
```diff
@@ -0,0 +1,301 @@
...
+  return htmlForEffect;
+}
+
+
+/* getNumberOfEffectParameter returns a number corresponding to the effect likelihood of 
+the keyframe image. This is used for the html meter which visually displays the likelihood 
+of each SafeSearch parameter on the page. The numbers returned are used to fill in the meter by 
+a certain amount (the number returned divided by 5).
+*/
+function getNumberOfEffectParameter(effectParameter) {
```
wdi-gg, line 20: Let's try this? https://google.github.io/styleguide/jsguide.html#jsdoc-enum-and-typedef-comments

capstone/src/main/webapp/js/image-effect-results.js
```diff
@@ -0,0 +1,301 @@
...
+  effectsAsNumbers.set(effect.violence, getNumberOfEffectParameter(effect.violence));
+  effectsAsNumbers.set(effect.racy, getNumberOfEffectParameter(effect.racy));
+
+  return effectsAsNumbers;
+}
+
+/* createKeyframeImageTextInnerHTML creates the html that displays the information about the keyframe image 
+that is displayed on the card shown to the user
+*/
+function createKeyframeImageTextInnerHTML(thisImage) {
```
wdi-gg, line 84: It's very clean since you move the HTML code into helper functions!

capstone/src/main/webapp/js/image-effect-results.js
```diff
@@ -0,0 +1,301 @@
...
+    keyframeImageDiv.appendChild(keyframeImage);
+
+    var imageCaptionDiv = document.createElement("div");
+    imageCaptionDiv.classList.add("container");
+
+    var effect = JSON.parse(thisImage.effect);
+
+    var effectsAsNumbers = setEffectsAsNumbers(effect);
+
+    // Don't display the image if it has no 4 or 5 (likely or very unlikely sensitive content),
```
wdi-gg, line 161: I think the behavior changed in the latest code?

capstone/src/main/webapp/js/image-effect-results.js
cassmarcussen: Fixed!

capstone/src/main/webapp/js/image-effect-results.js
cassmarcussen: Thanks!

capstone/src/main/java/com/google/sps/servlets/KeyframeImageUploadServlet.java
```diff
@@ -0,0 +1,142 @@
+package com.google.sps.servlets;
```
vkkpng-g, line 1: For this servlet we never check whether the fields are valid. I think Sharon also pointed this out (about time might be null). Are we sure that no matter what the fields (time, start, end) will always be valid?

capstone/src/main/java/com/google/sps/servlets/KeyframeImageUploadServlet.java
vkkpng-g: comment?

capstone/src/main/java/com/google/sps/servlets/KeyframeImageUploadServlet.java
vkkpng-g: comment?

capstone/src/main/webapp/css/style.css
```diff
@@ -1,5 +1,56 @@
...
+.navbar-dark .navbar-toggler {
+  border: 2px solid rgba(255, 255, 255, .6); 
+}
+
+.fa-bars {
+  color: rgba(255, 255, 255, .7);
+}
+
+.navbar .navbar-nav .nav-item .nav-link {
+  color: rgba(255, 255, 255, .7);
```
vkkpng-g, line 33: Curious, why do you have some colors define with rbga, some with hexadecimals value?

capstone/src/main/webapp/js/image-effect-results.js
```diff
@@ -0,0 +1,292 @@
...
+  return readableTimeFormatMinutesAndSeconds;
+}
+
+/* setEffectsAsNumbers sets the effectsAsNumbers map which has keys which are the effect categories and 
+values which are the numbers corresponding to the likelihood value for each of these effect categories
+*/
+function setEffectsAsNumbers(effect) {
+  var effectsAsNumbers = new Map();
+
+  effectsAsNumbers.set(effect.adult, getNumberOfEffectParameter(effect.adult));
```
vkkpng-g, line 63: I have no issue with how ur doing this 

+wdi do you think this is cleaner
var effect = [effect.adult, effect.medical, effect.spoofed, effect.violence, effect.racy]
for individualEffect in effect:
effectsAsNumbers.set(individualEffect getNumberOfEffectParameter(individualEffect));

capstone/src/main/webapp/js/image-effect-results.js
vkkpng-g: +1 yes, I noticed ur html code is very nice to understand and read.

capstone/src/main/webapp/js/image-effect-results.js
```diff
@@ -0,0 +1,292 @@
...
+  var startTime =  getReadableTimeFormat(thisImage.startTime);
+  var endTime =  getReadableTimeFormat(thisImage.endTime);
+
+  var keyframeImageTextInnerHTML = '<h2>Information about the frame</h2>'
+  + '<p>Timestamp of image: ' + timestamp + '</p>' 
+  + '<p>Start time of frame: ' + startTime + '</p>'
+  + '<p>End time of frame: ' + endTime + '</p>'
+  + '<hr>'
+  + '<h2>Effect of the frame </h2>' 
+  + htmlForEffect(effect.adult, effectsAsNumbers, "Adult")
```
vkkpng-g, line 87: I'm not html expert..but do u think its more readable if u extract the htmlForEffect to a var

var htmlForEffect = L87-91
Then in the code here all you have to do is "+ htmlForEffect"

capstone/src/main/webapp/js/image-effect-results.js
```diff
@@ -0,0 +1,292 @@
...
+  keyframeImageDiv.classList.add("mySlides");
+  keyframeImageDiv.classList.add("keyframe-card-fade");
+
+  var keyframeImage = document.createElement("img");
+  keyframeImage.src = thisImage.cloudBucketUrl.replace("gs://", "https://storage.cloud.google.com/");
+
+  // This condition makes sure the the keyframe image retrieved from the database is not undefined, 
+  // where undefined images either have a null src or 'undefined' in their source url. This is here 
+  // because we do not want to display undefined images (i.e. displaying no image) on the Results page.
+  if (keyframeImage.src != null && keyframeImage.src.indexOf("undefined") == -1) {
```
vkkpng-g, line 141: I think i made a comment above in ur servlet why you didn't check the image being null+other values. Is the reason u don't check for null value is because in the Javascript you check for validity?

If this is the case, we should check in the server code right? Like if its null, we don't even put the keyframe into the bucket. That way the javascript is just going to render the images.

capstone/src/main/webapp/js/image-effect-results.js
vkkpng-g: Ignore comment

capstone/src/main/webapp/js/image-effect-results.js
```diff
@@ -0,0 +1,292 @@
...
+  showSlides(slideIndex += n);
+}
+
+// Thumbnail image controls
+function currentSlide(n) {
+  showSlides(slideIndex = n);
+}
+
+function showSlides(n) {
+  var i;
```
vkkpng-g, line 278: don't use naming like "i". Is this the indexToSlides?

capstone/src/main/webapp/js/image-effect-results.js
```diff
@@ -0,0 +1,292 @@
...
+
+/* Based on https://www.w3schools.com/howto/howto_js_slideshow.asp */
+
+// Next/previous controls
+function plusSlides(n) {
+  showSlides(slideIndex += n);
+}
+
+// Thumbnail image controls
+function currentSlide(n) {
```
vkkpng-g, line 273: Don't use "n" choose a better name representative of what n is

capstone/src/main/webapp/js/image-effect-results.js
```diff
@@ -0,0 +1,292 @@
...
+
+  /* location.reload() is a predefined JS function for a predefined JS class named location. */
+  location.reload();
+}
+
+
+/* Based on https://www.w3schools.com/howto/howto_js_slideshow.asp */
+
+// Next/previous controls
+function plusSlides(n) {
```
vkkpng-g, line 268: Don't use "n" choose a better name representative of what n is

capstone/src/main/webapp/js/image-effect-results.js
```diff
@@ -0,0 +1,292 @@
+var slideIndex = 1;
+
+window.onload = function() {
+  fetchBlobstoreKeyframeImages();
+};
+
+function htmlForEffect(effectForACategory, effectsAsNumbers, categoryName) {
+  var htmlForEffect = '<p><label for="adult">' + categoryName + ': ';
```
vkkpng-g, line 8: why do we name this "adult" in the "label for"?

capstone/src/main/webapp/results.html
```diff
@@ -1,14 +1,71 @@
...
+            <!-- Next and previous buttons -->
+            <a class="prev" onclick="plusSlides(-1)">&#10094;</a>
+            <a class="next" onclick="plusSlides(1)">&#10095;</a>
+        </div>
+        <br>
+
+        <!-- The dots/circles -->
+        <div id="dots" style="text-align:center"> </div>
+
+        <!--UNCOMMENT TO TEST THE DELETEENTRIES FUNCTIONALITY
```
vkkpng-g, line 66: will this code be deleted?

capstone/src/main/webapp/css/style.css
cassmarcussen: Good point! There's not really a reason, so it would be good to be consistent.

capstone/src/main/webapp/js/image-effect-results.js
cassmarcussen: Fixed

capstone/src/main/webapp/results.html
cassmarcussen: Yes, at some point. For now it is still useful in testing my component, but I'll delete it once we have the delete incorporated into the merged copy on master.

capstone/src/main/webapp/js/image-effect-results.js
cassmarcussen: Fixed

capstone/src/main/webapp/js/image-effect-results.js
cassmarcussen: Fixed

capstone/src/main/webapp/js/image-effect-results.js
cassmarcussen: Fixed

capstone/src/main/webapp/js/image-effect-results.js
cassmarcussen: Based on [documentation](https://www.w3schools.com/tags/att_label_for.asp) I've read for meter labels, the syntax for labels for meters is: \<label for="element_id"\>

capstone/src/main/java/com/google/sps/servlets/KeyframeImageUploadServlet.java
cassmarcussen: Fixed

capstone/src/main/webapp/js/image-effect-results.js
cassmarcussen: Moved value checking to server side

capstone/src/main/java/com/google/sps/servlets/KeyframeImageUploadServlet.java
cassmarcussen: No longer using / storing start and end time

capstone/src/main/java/com/google/sps/data/KeyframeImage.java
cassmarcussen: Fixed

capstone/src/main/java/com/google/sps/data/KeyframeImage.java
cassmarcussen: Renamed safeSearchEffect

capstone/src/main/java/com/google/sps/servlets/KeyframeImageDeleteServlet.java
cassmarcussen: Fixed

capstone/src/main/java/com/google/sps/servlets/KeyframeImageDeleteServlet.java
cassmarcussen: Fixed

capstone/src/main/java/com/google/sps/servlets/KeyframeImageUploadServlet.java
cassmarcussen: Fixed

capstone/src/main/java/com/google/sps/servlets/KeyframeImageUploadServlet.java
cassmarcussen: Added checks for url and timestamp. Start and end times are now removed.

capstone/src/main/webapp/js/image-effect-results.js
cassmarcussen: Changed to switch case, now using an enum based on a suggestion made by Di

capstone/src/main/webapp/js/image-effect-results.js
cassmarcussen: Made a comment, but removed comment now and added this check to the server side (upload servlet)

capstone/src/main/webapp/js/image-effect-results.js
cassmarcussen: Fixed

capstone/src/main/webapp/js/image-effect-results.js
cassmarcussen: Fixed

capstone/src/main/java/com/google/sps/servlets/DetectSafeSearchGcs.java
```diff
@@ -0,0 +1,79 @@
...
+  See the function below for a description. This function calls the one below, passing in the image's url as a parameter. */
+  public static HashMap<String, String> detectSafeSearchGcs(KeyframeImage image) throws IOException {
+    String filePath = image.getUrl();
+    return detectSafeSearchGcs(filePath);
+  }
+
+  /* Detects whether the specified image on Google Cloud Storage has features you would want to moderate. */
+  public static HashMap<String, String> detectSafeSearchGcs(String gcsPath) throws IOException {
+
+    HashMap<String, String> safeSearchResults = new HashMap<String, String>();
```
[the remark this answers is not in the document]
cassmarcussen: Fixed

capstone/src/main/java/com/google/sps/servlets/DetectSafeSearchGcs.java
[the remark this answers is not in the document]
cassmarcussen: Fixed

capstone/src/main/java/com/google/sps/servlets/DetectSafeSearchGcs.java
```diff
@@ -0,0 +1,79 @@
...
+
+        // For full list of available annotations, see http://g.co/cloud/vision/docs
+        SafeSearchAnnotation annotation = res.getSafeSearchAnnotation();
+
+        safeSearchResults.put("adult", annotation.getAdult().toString());
+        safeSearchResults.put("medical", annotation.getMedical().toString());
+        safeSearchResults.put("spoofed", annotation.getSpoof().toString());
+        safeSearchResults.put("violence", annotation.getViolence().toString());
+        safeSearchResults.put("racy", annotation.getRacy().toString());
+
```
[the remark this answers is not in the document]
cassmarcussen: Fixed

capstone/src/main/java/com/google/sps/servlets/DetectSafeSearchGcs.java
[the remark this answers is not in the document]
cassmarcussen: Fixed

capstone/src/main/java/com/google/sps/servlets/DetectSafeSearchGcs.java
```diff
@@ -0,0 +1,79 @@
...
+      for (AnnotateImageResponse res : responses) {
+        if (res.hasError()) {
+          System.out.format("Error: %s%n", res.getError().getMessage());
+          safeSearchResults.put("adult", res.getError().getMessage());
+          safeSearchResults.put("medical", "UNKNOWN");
+          safeSearchResults.put("spoofed", "UNKNOWN");
+          safeSearchResults.put("violence", "UNKNOWN");
+          safeSearchResults.put("racy", "UNKNOWN");
+          return safeSearchResults;
```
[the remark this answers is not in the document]
cassmarcussen: Fixed

capstone/src/main/java/com/google/sps/servlets/KeyframeImageDeleteServlet.java
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+  }
+
+  /* Deletes all objects from the Google Cloud Bucket associated with the keyframe images for the video.
+  Reference: https://cloud.google.com/storage/docs/deleting-objects#storage-delete-object-java 
+  */
+  private void deleteGoogleCloudBucketInfo() {
+    // The ID of your GCP project
+    String projectId = "video-vigilance";
```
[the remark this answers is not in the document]
cassmarcussen: Fixed

capstone/src/main/java/com/google/sps/servlets/KeyframeImageUploadServlet.java
cassmarcussen: Fixed

capstone/src/main/webapp/css/style.css
cassmarcussen: Note: I will change this in another branch, that has more CSS, so that making this change is simpler.

capstone/src/main/webapp/js/image-effect-results.js
cassmarcussen: Fixed

capstone/src/main/webapp/js/image-effect-results.js
cassmarcussen: Fixed

PR #6, opened 2020-07-01
Extracting image frames

sharond106:
Note: I accidentally merged some of Cassandra's code into this branch, but I reversed it so this PR only contains my code. Her commits are just still showing.

**1. Use Video Intelligence API's shot detection feature to detect shot changes in a video**

I created a java servlet (ShotsServlet.java) that sends a request to the API for the shot start and end times and outputs these times as a json object. Currently, a hard coded url to a Google Cloud bucket that contains a video file is being used. A javascript function (getShots) sends a request to the ShotsServlet and parses the shot times from the json object. 

Retrieving shots from servlet: https://screenshot.googleplex.com/vh9qGAqprqo
Received shot times: https://screenshot.googleplex.com/4mTSFwmFKxp

**2. Use javascript to extract image frames from a video and display them on the page (DIAGRAM: https://screenshot.googleplex.com/t7FA3WkAONA)**

Currently, the javascript code is reading a video uploaded to the webpage by the user, so I have been testing my code by uploading the same video that's in the hard coded bucket url. I couldn't pass in the bucket url because I was running into permission errors.

Upload a video file and show images: https://screenshot.googleplex.com/w3z1DkpLAkk, https://screenshot.googleplex.com/Dzde4U9D5W1


capstone/src/main/webapp/js/extract-images-script.js
```diff
@@ -12,3 +12,154 @@
...
+// Array of times to keyframe images at
+const keyTimes = [];
+
+// Current index of keyTimes
+var keyTimesIndex = 0;
+
+/** 
+ * Gets the first frame in the video by calling captureFrame
+ */ 
+function firstFrame() {
```
rwackes, line 26: Change order of functions from top to bottom to match how you are calling them in index.html

capstone/src/main/webapp/js/extract-images-script.js
```diff
@@ -12,3 +12,154 @@
...
+/** 
+ * Draws a frame of the video onto a canvas element
+ * 
+ * @param {string} path: The path of the video file
+ * @param {number} secs: The time of the video frame to be captured in seconds
+ */
+function captureFrame(path, secs) {
+
+	// Load video src (needs to be reloaded for events to be triggered)
+	const video = document.getElementById("video");
```
rwackes, line 55: fix indentation

capstone/src/main/webapp/js/extract-images-script.js
```diff
@@ -12,3 +12,154 @@
...
+  video.src = path;
+
+	// When the metadata has been loaded, set the time of the video to be captured
+	video.onloadedmetadata = function() {
```
rwackes, line 59: fix indentation

capstone/src/main/webapp/js/extract-images-script.js
```diff
@@ -12,3 +12,154 @@
...
+
+      keyTimes.push((shot.start_time + shot.end_time) / 2.0);
+      count++;
+    }
+  });
+}
+
+// Displays the video to the webpage
+function showVideo() {
+	const video = document.getElementById("video");
```
rwackes, line 156: fix indentation to two spaces and not tab

capstone/src/main/webapp/js/extract-images-script.js
```diff
@@ -12,3 +12,154 @@
...
+	video.src = URL.createObjectURL(document.querySelector("#video-file").files[0]);
+	video.style.display = "block";
+}
+
+// Hides the video from the webpage
+function hideVideo() {
+	const video = document.getElementById("video");
```
rwackes, line 163: fix indentation to two spaces and not tab

capstone/src/main/java/com/google/sps/servlets/ShotsServlet.java
```diff
@@ -0,0 +1,99 @@
...
+    String json = gson.toJson(shots);
+    response.setContentType("application/json;");
+    response.getWriter().println(json);
+  }
+
+  // Performs shot analysis on the video at the provided Cloud Storage path
+  private void detectShots(String gcsUri) throws Exception {
+
+    // Instantiate a com.google.cloud.videointelligence.v1.VideoIntelligenceServiceClient
+    try (VideoIntelligenceServiceClient client = VideoIntelligenceServiceClient.create()) {
```
cassmarcussen, line 68: Maybe move the code from inside the parameter to inside the body of the try statement. From my understanding of try/catch in Java, the program tries the entire code in the 'try', and throws an exception if any of these lines cause an issue, so moving this code to the try body should catch exceptions.

Also, does this try have a catch associated with it?

I saw this same line of code and made this comment in Rhiannan's code review, so I am thinking that this is code based on the Video Intelligence walkthroughs, but it may still be worthwhile to make this change.

capstone/src/main/java/com/google/sps/servlets/ShotsServlet.java
```diff
@@ -0,0 +1,99 @@
...
+          .build();
+ 
+      // Create an operation that will contain the response when the operation completes.
+      OperationFuture<AnnotateVideoResponse, AnnotateVideoProgress> response =
+          client.annotateVideoAsync(request);
+
+      // Get annotations results for each video sent (we will only be sending 1 video)
+      for (VideoAnnotationResults result : response.get().getAnnotationResultsList()) {
+        if (result.getShotAnnotationsCount() > 0) {
+
```
cassmarcussen, line 83: Remove extra blank line. In this file and similar files, make sure your spacing / use of blank lines is consistent, for readibility.

capstone/src/main/java/com/google/sps/servlets/ShotsServlet.java
```diff
@@ -0,0 +1,99 @@
...
+          // Get shot annotations for video
+          for (VideoSegment segment : result.getShotAnnotationsList()) {
+            double startTime = segment.getStartTimeOffset().getSeconds()
+                + segment.getStartTimeOffset().getNanos() / 1e9;
```
cassmarcussen, line 87: Add a comment about what 1e9 is, since it isn't immediately clear what its purpose is.

capstone/src/main/webapp/css/style.css
```diff
@@ -1,5 +1,9 @@
 #content {
-  margin-left: auto;
-  margin-right: auto;
+  margin: 80px;
   width: 650px;
 }
+
+#video,
+#video-loaded {
```
cassmarcussen, line 7: Put #video and #video-loaded on the same line

capstone/src/main/webapp/index.html
```diff
@@ -3,12 +3,27 @@
...
   <body>
     <div id="content">
-      <h1>Video Vigilance</h1>
+      <button onclick="getShots()">Detect shots</button>
+      <br>
+      <ul id="shots-list"></ul>
+      <p id="loading"></p>
+      <hr>
+      <input type="file" id="video-file" name="video-file" accept="video/*">
+      <br>
```
cassmarcussen, line 17: Instead of using one or more consecutive break tags in your html, could you move the code that you want to be spaced out into divs that have padding and/or margins specified?

capstone/src/main/webapp/js/extract-images-script.js
```diff
@@ -12,3 +12,154 @@
...
+var keyTimesIndex = 0;
+
+/** 
+ * Gets the first frame in the video by calling captureFrame
+ */ 
+function firstFrame() {
+	
+	// If there are no shots to display, show error message
+	if (keyTimes.length == 0) {
+		const li = document.createElement("li");
```
cassmarcussen, line 30: Too much indentation (use 2 spaces to indent)

capstone/src/main/webapp/js/extract-images-script.js
```diff
@@ -12,3 +12,154 @@
...
+		li.innerHTML += "No shots to display<br>";
+		document.getElementById("frames-list").appendChild(li);
+		return;
+	} 
+	// Otherwise, initialize variables
+	else {
+		keyTimesIndex = 0;
```
cassmarcussen, line 37: Too much indentation

capstone/src/main/webapp/js/extract-images-script.js
```diff
@@ -12,3 +12,154 @@
...
+ * Draws a frame of the video onto a canvas element
+ * 
+ * @param {string} path: The path of the video file
+ * @param {number} secs: The time of the video frame to be captured in seconds
+ */
+function captureFrame(path, secs) {
+
+	// Load video src (needs to be reloaded for events to be triggered)
+	const video = document.getElementById("video");
+  video.src = path;
```
cassmarcussen, line 56: Indent this line

capstone/src/main/webapp/js/extract-images-script.js
cassmarcussen, line 59: Indentation (make sure the contents of the function are more indented than the first line, and make sure the opening and closing curly brackets are evenly indented with each other).

capstone/src/main/webapp/js/extract-images-script.js
```diff
@@ -12,3 +12,154 @@
...
+
+	// When the metadata has been loaded, set the time of the video to be captured
+	video.onloadedmetadata = function() {
+    this.currentTime = secs;
+  };
+	
+	// When the video has seeked to the specific time, draw the frame onto a canvas element
```
cassmarcussen, line 63: For the rest of the file, fix the indentation of some of the comments (they should be indented as much as the lines they comment about, not more indented)

capstone/src/main/webapp/js/extract-images-script.js
```diff
@@ -12,3 +12,154 @@
...
+	else {
+		li.innerHTML += "Error capturing frame";
+	}
+
+	document.getElementById("frames-list").appendChild(li);
+
+	// Check if there are more frames to capture
+	if (++keyTimesIndex < keyTimes.length) {
+		captureFrame(
+			URL.createObjectURL(document.querySelector("#video-file").files[0]),
```
cassmarcussen, line 116: Line 108, 116: too much indentation. Use 2 spaces for indentation.

capstone/src/main/java/com/google/sps/servlets/ShotsServlet.java
sharond106: Ok! It's just getting the nanoseconds and adding it onto the seconds ('54' nanoseconds --> decimal form of nanoseconds).

capstone/src/main/java/com/google/sps/servlets/ShotsServlet.java
sharond106: I tend to always start comments with a blank line before for some reason. I'll try to keep for loops and if/else blocks together so it's more readable!

capstone/src/main/java/com/google/sps/servlets/ShotsServlet.java
sharond106: Oh yeah I think it will work the same. It doesn't have a catch associated with it because the function throws Exception. I catch for it in the doGet method when I call this detectShots method

capstone/src/main/webapp/index.html
sharond106: yep!

capstone/src/main/java/com/google/sps/servlets/ShotsServlet.java
```diff
@@ -0,0 +1,99 @@
...
+
+    // Using hard coded image file for now
+    String gcsUri = "gs://video-vigilance-videos/youtube_ad_test.mp4";
+
+    // Get detected shot times
+    try {
+      detectShots(gcsUri);
+    } catch (Exception e) {
+
+      // TODO: do something else here that's meaningful on the client side, not sure what to do yet
```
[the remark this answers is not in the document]
sharond106: Yeah good thinking. I was thinking of adding a json object to the json response array. So the final json response would be like: "[exception: false, shots array list]". And on the client side, check if "exception" is true before parsing the shots array list. Any thoughts/ other ideas?

Also, are our processes overlapping? I thought I only affected Cassandra's code.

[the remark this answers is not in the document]
rwackes: I just need to know how you are keeping/storing the video file that the user submits. I need to use the gcs path to it for my stuff.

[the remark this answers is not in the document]
sharond106: oh ok. I'm just going to leave the error here for now because in my other branch, I handle the case where the returned shots array is empty

capstone/src/main/webapp/css/style.css
```diff
@@ -3,7 +3,10 @@
...
 }
 
-#video,
-#video-loaded {
+#video, #video-loaded {
   display: none;
 }
+
+#frames-list, #video {
+  padding: 50px, 0;
```
cassmarcussen, line 11: Maybe add a 'px' after the 0, so that it is consistent with the 50px

capstone/src/main/java/com/google/sps/servlets/ShotsServlet.java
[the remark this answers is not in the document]
rwackes: yes your idea sounds good.

capstone/src/main/java/com/google/sps/data/Shot.java
```diff
@@ -0,0 +1,26 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.sps.data;
+
+/** A detected shot, with start and end time offsets (seconds)**/
```
rwackes, line 17: small thing but end the comment with just one * like */ and not **/

capstone/src/main/java/com/google/sps/servlets/ShotsServlet.java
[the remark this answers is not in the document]
rwackes: similarly, since i am returning a HashMap, I added an "error" key and check for it in my JS to determine what to display to the user/what code to execute/not execute bc of the error.

capstone/src/main/java/com/google/sps/servlets/ShotsServlet.java
```diff
@@ -0,0 +1,96 @@
...
+    String gcsUri = "gs://video-vigilance-videos/youtube_ad_test.mp4";
+
+    // Get detected shot times
+    try {
+      detectShots(gcsUri);
+    } catch (Exception e) {
+      e.printStackTrace(response.getWriter());
+    }
+
+    // Create json String with shots objects
```
vkkpng-g, line 55: Hmm so in the code below you are creating shots even though there might be an error. Is this expected?

capstone/src/main/java/com/google/sps/servlets/ShotsServlet.java
```diff
@@ -0,0 +1,96 @@
...
+    Gson gson = new Gson();
+    String json = gson.toJson(shots);
```
wdi-gg, line 57: Maybe add a comment of a sample output?

capstone/src/main/java/com/google/sps/servlets/ShotsServlet.java
vkkpng-g: It would be good to add a comment that another method (name which one) does the catching. Was about to make same comment

capstone/src/main/java/com/google/sps/servlets/ShotsServlet.java
```diff
@@ -0,0 +1,96 @@
...
+      AnnotateVideoRequest request = AnnotateVideoRequest.newBuilder()
+          .setInputUri(gcsUri)
+          .addFeatures(Feature.SHOT_CHANGE_DETECTION)
+          .build();
+ 
+      // Create an operation that will contain the response when the operation completes.
+      OperationFuture<AnnotateVideoResponse, AnnotateVideoProgress> response =
+          client.annotateVideoAsync(request);
+
+      // Get annotations results for each video sent (we will only be sending 1 video)
```
wdi-gg, line 78: Does this mean "response.get().getAnnotationResultsList()" should return a list of size 1?

capstone/src/main/java/com/google/sps/servlets/ShotsServlet.java
```diff
@@ -0,0 +1,96 @@
...
+      for (VideoAnnotationResults result : response.get().getAnnotationResultsList()) {
+        if (result.getShotAnnotationsCount() > 0) {
```
wdi-gg, line 80: It's great that we are being cautious! But I'm not sure about its effect, i.e. without the check, will we also just skip the for loop when count is 0?

capstone/src/main/java/com/google/sps/servlets/ShotsServlet.java
```diff
@@ -0,0 +1,96 @@
...
+          // Get shot annotations for video
+          for (VideoSegment segment : result.getShotAnnotationsList()) {
```
wdi-gg, line 82: How many segments will be returned for a video?

capstone/src/main/java/com/google/sps/servlets/ShotsServlet.java
```diff
@@ -0,0 +1,96 @@
...
+  // Performs shot analysis on the video at the provided Cloud Storage path
+  private void detectShots(String gcsUri) throws Exception {
+      
+    // Instantiate a com.google.cloud.videointelligence.v1.VideoIntelligenceServiceClient
+    try (VideoIntelligenceServiceClient client = VideoIntelligenceServiceClient.create()) {
+        
+      // Provide path to file hosted on GCS as "gs://bucket-name/..."
+      AnnotateVideoRequest request = AnnotateVideoRequest.newBuilder()
+          .setInputUri(gcsUri)
+          .addFeatures(Feature.SHOT_CHANGE_DETECTION)
```
vkkpng-g, line 71: Add comment what shot_change_detection is

capstone/src/main/java/com/google/sps/servlets/ShotsServlet.java
vkkpng-g, line 80: Instead of inner if loops

if (result.getShotAnnotationCount() = 0)
continue

capstone/src/main/java/com/google/sps/servlets/ShotsServlet.java
```diff
@@ -0,0 +1,96 @@
...
+      // Get annotations results for each video sent (we will only be sending 1 video)
+      for (VideoAnnotationResults result : response.get().getAnnotationResultsList()) {
+        if (result.getShotAnnotationsCount() > 0) {
+          // Get shot annotations for video
+          for (VideoSegment segment : result.getShotAnnotationsList()) {
+            // Add on nanoseconds to total seconds
+            double startTime = segment.getStartTimeOffset().getSeconds()
+                + segment.getStartTimeOffset().getNanos() / 1e9;
+            double endTime = segment.getEndTimeOffset().getSeconds()
+                + segment.getEndTimeOffset().getNanos() / 1e9;
```
vkkpng-g, line 87: In you Design perspective should we round the nunber to just seconds. Like I looks at ur screenshot and i saw 12.1213212312. I don't think the values after the decimal have much value since users can't exactly go to that time frame manually?

capstone/src/main/webapp/js/extract-images-script.js
```diff
@@ -12,3 +12,147 @@
...
+      list.append(listElement);
+      keyTimes.push((shot.start_time + shot.end_time) / 2.0);
+      count++;
+    }
+  });
+}
+
+// Gets the first frame in the video by calling captureFrame
+function firstFrame() {
+  // If there are no shots to display, show error message
```
wdi-gg, line 54: We can do the following to comment on if-else.
if () {
// comments
} else {
// comments
}

capstone/src/main/webapp/js/extract-images-script.js
```diff
@@ -12,3 +12,147 @@
...
+      listElement.appendChild(textElement);
+      list.append(listElement);
+      keyTimes.push((shot.start_time + shot.end_time) / 2.0);
+      count++;
+    }
+  });
+}
+
+// Gets the first frame in the video by calling captureFrame
+function firstFrame() {
```
wdi-gg, line 53: Does this function capture only the first frame or all the frames?

capstone/src/main/webapp/js/extract-images-script.js
```diff
@@ -12,3 +12,147 @@
...
+  } 
+  // If the video was not successfully seeked, display error message
+  else {
+    li.innerHTML += "Error capturing frame";
+  }
+
+  document.getElementById("frames-list").appendChild(li);
+
+  // Check if there are more frames to capture
+  if (++keyTimesIndex < keyTimes.length) {
```
wdi-gg, line 139: Is there a reason that we are doing this, rather than just do a for loop at line 66?

capstone/src/main/webapp/js/extract-images-script.js
```diff
@@ -12,3 +12,147 @@
...
+ * @param {number} secs: The time of the video frame to be captured in seconds
+ */
+function captureFrame(path, secs) {
+  // Load video src (needs to be reloaded for events to be triggered)
+  const video = document.getElementById("video");
+  video.src = path;
+
+  // When the metadata has been loaded, set the time of the video to be captured
+  video.onloadedmetadata = function() {
+    this.currentTime = secs;
```
wdi-gg, line 85: Is there a reason that we need to assign secs to currentTime, rather than just pass secs in line 107?

capstone/src/main/webapp/js/extract-images-script.js
```diff
@@ -12,3 +12,147 @@
...
+  };
+	
+  // When the video has seeked to the specific time, draw the frame onto a canvas element
+  video.onseeked = function(event) {
+    const canvas = document.createElement("canvas");
+    canvas.height = video.videoHeight;
```
wdi-gg, line 91: Is there a reason that we need to assign height / width, rather than just pass them in line 100?

capstone/src/main/webapp/js/extract-images-script.js
```diff
@@ -12,3 +12,147 @@
...
+    var count = 1;
+
+    // Display each shot's times in a list and add the middle time of each shot to keyTimes array
+    for (const shot of shots) {
+      const listElement = document.createElement("li");
+      const textElement = document.createElement("span");
+      textElement.innerHTML = "<b>Shot " + count + ": <b>" + shot.start_time + " - " + shot.end_time;
+      listElement.appendChild(textElement);
+      list.append(listElement);
+      keyTimes.push((shot.start_time + shot.end_time) / 2.0);
```
vkkpng-g, line 46: What is keyTimes

capstone/src/main/webapp/js/extract-images-script.js
```diff
@@ -12,3 +12,147 @@
...
+
+  fetch("/shots").then(response => response.json()).then(shots => {
+    // Remove loading message
+    const message = document.getElementById("loading");
+    message.innerHTML = "";
+
+    // Display shot times to user
+    const list = document.getElementById("shots-list");
+    list.innerHTML = "";
+    var count = 1;
```
vkkpng-g, line 37: what is this count for?

capstone/src/main/webapp/js/extract-images-script.js
```diff
@@ -12,3 +12,147 @@
...
+    document.getElementById("frames-list").appendChild(li);
+    return;
+  } 
+  // Otherwise, initialize variables
+  else {
+    keyTimesIndex = 0;
+    document.getElementById("frames-list").innerHTML = "";
+  }
+  captureFrame(
+    URL.createObjectURL(document.querySelector("#video-file").files[0]),
```
wdi-gg, line 67: Make sure files[0] won't overflow?

capstone/src/main/webapp/js/extract-images-script.js
```diff
@@ -12,3 +12,147 @@
...
+  };
+	
+  // When the video has seeked to the specific time, draw the frame onto a canvas element
+  video.onseeked = function(event) {
+    const canvas = document.createElement("canvas");
+    canvas.height = video.videoHeight;
+    canvas.width = video.videoWidth;
+
+    // Get 2d drawing context on canvas
+    const ctx = canvas.getContext("2d");
```
vkkpng-g, line 95: what does ctx mean?

capstone/src/main/webapp/js/extract-images-script.js
```diff
@@ -12,3 +12,147 @@
...
+  // If the video was not successfully seeked, display error message
+  else {
+    li.innerHTML += "Error capturing frame";
+  }
+
+  document.getElementById("frames-list").appendChild(li);
+
+  // Check if there are more frames to capture
+  if (++keyTimesIndex < keyTimes.length) {
+    captureFrame(
```
vkkpng-g, line 140: For the html to work do we have to create each one like this. Its very confusing why we are calling captureFram after creating the html and doing it for each frame. ideally it would be good to do

html string
for each frame:
   document.getElementById("frames-list").appendChild(li);

return html string

capstone/src/main/java/com/google/sps/servlets/ShotsServlet.java
sharond106: The code below is just creating a json object of the shots array list. Since the array list is initialized beforehand, it should just be an empty json array, which is expected.

capstone/src/main/java/com/google/sps/servlets/ShotsServlet.java
sharond106: Sure! Added on line 55

capstone/src/main/java/com/google/sps/servlets/ShotsServlet.java
sharond106: Added on line 64

capstone/src/main/java/com/google/sps/servlets/ShotsServlet.java
sharond106: Added line 73

capstone/src/main/java/com/google/sps/servlets/ShotsServlet.java
sharond106: In our project, since we are only analyzing 1 video at once, yes it should just return a list of size 1.

capstone/src/main/java/com/google/sps/servlets/ShotsServlet.java
sharond106: Yes, the for loop will be skipped, and the function will not add any new Shot objects to the shots array list (an empty shots list will be returned by the servlet).

sharond106: Fixed!

capstone/src/main/java/com/google/sps/servlets/ShotsServlet.java
sharond106: Segment is just the java variable name for "shot". So it depends how many shots are detected in the video

capstone/src/main/java/com/google/sps/servlets/ShotsServlet.java
sharond106: That's a good point. I'll change them to ints.

capstone/src/main/webapp/js/extract-images-script.js
sharond106: It's for the displayed shot list labels. In line 43 I'm using it as a counter for each label like: "Shot x:". I think the UI will be changing later, so this variable may not be needed in the future.

capstone/src/main/webapp/js/extract-images-script.js
sharond106: I'm capturing 1 frame per shot, so I'm adding the middle time of each shot to the keyTimes array. The middle time is calculated by averaging the start and end times.

capstone/src/main/webapp/js/extract-images-script.js
sharond106: This function does all the checks/ initializing necessary before all the shot capturing happens. If the API was able to return shot times, then it will call the captureFrame function, which will start capturing frames.

capstone/src/main/webapp/js/extract-images-script.js
sharond106: Fixed!

capstone/src/main/webapp/js/extract-images-script.js
sharond106: I created a global variable for the file path and add a check in firstFrame for it. It will also get set in firstFrame.

capstone/src/main/webapp/js/extract-images-script.js
sharond106: currentTime is a property of the "video" variable. It's setting the video's to the time that we want to capture a frame at

capstone/src/main/webapp/js/extract-images-script.js
sharond106: Lines 91 + 92 are setting the dimensions of the HTML canvas element that we are drawing the video frame onto. It must be set for proper scaling of the drawn image

capstone/src/main/webapp/js/extract-images-script.js
sharond106: ctx is the variable containing the 2d drawing context of the canvas. It's how a canvas element can be drawn on

capstone/src/main/webapp/js/extract-images-script.js
sharond106: Yes. I initially tried doing it with a for loop, but javascript being synchronous would not wait for iterations of the loop to finish before moving on to the next iteration. With the for loop, only the very last shot time would be captured and displayed. This method of using callback functions ensures that the entire process of capturing 1 image frame will be finished before moving on to the next shot time to capture.

capstone/src/main/webapp/js/extract-images-script.js
sharond106: I think this is related to Di's comment about for loops, but I was having trouble with getting a for loop to work. I can explain it more in person, but would you prefer if I tried doing it without these callback functions instead?

capstone/src/main/java/com/google/sps/servlets/ShotsServlet.java
sharond106: I changed it back to the decimals- I'm going to capture the frames at the decimal times, but display a rounded number to the user

capstone/src/main/java/com/google/sps/servlets/ShotsServlet.java
wdi-gg: Do we want to handle the case when it's not 1?

capstone/src/main/java/com/google/sps/servlets/ShotsServlet.java
```diff
@@ -0,0 +1,100 @@
...
+          .addFeatures(Feature.SHOT_CHANGE_DETECTION)
+          .build();
+ 
+      // Create an operation that will contain the response when the operation completes.
+      OperationFuture<AnnotateVideoResponse, AnnotateVideoProgress> response =
+          client.annotateVideoAsync(request);
+
+      // Get annotations results for each video sent (we will only be sending 1 video)
+      for (VideoAnnotationResults result : response.get().getAnnotationResultsList()) {
+        if (result.getShotAnnotationsCount() == 0) {
```
wdi-gg, line 83: Is this line necessary?

capstone/src/main/webapp/js/extract-images-script.js
wdi-gg: This is a good fallback option. Maybe explore the for loop option later if you have time :)

capstone/src/main/webapp/js/extract-images-script.js
wdi-gg: Thanks for the diagram!
We might re-arrange the structure later if we explore the for loop option. This is fine for now, and maybe just change it to InitializeFrame or something, since reader might find current name confusing and wonder what's so special about the first frame?

capstone/src/main/webapp/js/extract-images-script.js
wdi-gg: This is a good fallback option. Maybe explore for loop option later when you have time :)

capstone/src/main/java/com/google/sps/servlets/ShotsServlet.java
sharond106: Ok. I changed it so that only the first one will be looked at

capstone/src/main/webapp/js/extract-images-script.js
sharond106: Ok sounds good!

capstone/src/main/java/com/google/sps/servlets/ShotsServlet.java
sharond106: hm I guess not because the for loop below it should just get skipped in the case there are no annotations! I deleted it

PR #7, opened 2020-07-06
Adding junit tests for shots

sharond106:
Create junit tests for DetectShotsServlet.java

I moved the code in DetectShotsServlet.java to DetectShots.java so that the method can be tested from ShotDetectionTest.java without sending HTTP requests. The API calls are in a helper function in DetectShots.java, which I stub in my tests.

Test results: https://screenshot.googleplex.com/wQFGN6qKsHF


sharond106:
I'm still confused how to really test the connection to the API. I think lines 37-41 in DetectShots.java should run properly as long as the gcsUri is valid. But everything after these lines will give a permission denied error running locally. Since I can't mock private methods, is the only way to test these lines to move it to a public method? I didn't want to do this because it messes up the pipeline

capstone/src/test/java/com/google/sps/ShotDetectionTest.java
```diff
@@ -0,0 +1,82 @@
...
+import org.junit.runners.JUnit4;
+
+import static org.mockito.Mockito.*;
+import static org.mockito.Matchers.*;
+
+
+@RunWith(JUnit4.class)
+public final class ShotDetectionTest {
+
+  private DetectShots detectShots = new DetectShots();
```
rwackes, line 38: follow the guidelines under the "Setup" section in this link and move initialization code into a setup function using the @Before annotation: https://happycoding.io/tutorials/java/unit-testing

capstone/src/test/java/com/google/sps/ShotDetectionTest.java
```diff
@@ -0,0 +1,82 @@
...
+  private DetectShots mockDetectShots = mock(DetectShots.class);
```
rwackes, line 39: create mock using Mockito.mock() and not mock() according to https://happycoding.io/tutorials/java/unit-testing

capstone/src/test/java/com/google/sps/ShotDetectionTest.java
```diff
@@ -25,35 +28,44 @@
 import org.junit.runner.RunWith;
 import org.junit.runners.JUnit4;
 
-/** */
 @RunWith(JUnit4.class)
 public final class ShotDetectionTest {
```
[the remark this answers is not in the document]
sharond106: Done!

capstone/src/test/java/com/google/sps/ShotDetectionTest.java
sharond106: Done!

capstone/src/test/java/com/google/sps/ShotDetectionTest.java
sharond106: I think this is the same thing I'm doing since I imported Mockito.* but that article only imported up to Mockito..?

capstone/src/main/java/com/google/sps/data/DetectShots.java
```diff
@@ -0,0 +1,69 @@
...
+ 
+      // Create an operation that will contain the response when the operation completes.
+      OperationFuture<AnnotateVideoResponse, AnnotateVideoProgress> response =
+          client.annotateVideoAsync(request);
+    
+      // Instantiate list to add shots to
+      ArrayList<Shot> shots = new ArrayList<Shot>();
+
+      // Get annotations results for each video sent (we will only be sending 1 video)
+      for (VideoAnnotationResults result : response.get().getAnnotationResultsList()) {
```
vkkpng-g, line 51: Add test 
1) When the list is empty

capstone/src/main/java/com/google/sps/data/DetectShots.java
```diff
@@ -0,0 +1,69 @@
...
+          // Get shot annotations for video
+          for (VideoSegment segment : result.getShotAnnotationsList()) {
+            // Add on nanoseconds to total seconds
+            double startTime = segment.getStartTimeOffset().getSeconds()
+                + segment.getStartTimeOffset().getNanos() / 1e9;
+            double endTime = segment.getEndTimeOffset().getSeconds()
+                + segment.getEndTimeOffset().getNanos() / 1e9;
+            // Create Shot object and add to shots ArrayList
+            Shot newShot = new Shot(startTime, endTime);
+            shots.add(newShot);
```
vkkpng-g, line 62: I believe your test only deals with 1 shot. Good to have test that deals with more (like 3?)

capstone/src/main/java/com/google/sps/servlets/ShotsServlet.java
```diff
@@ -47,7 +40,8 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
 
     // Get detected shot times
     try {
-      detectShots(gcsUri);
+      DetectShots detectShots = new DetectShots();
+      shots = detectShots.detect(gcsUri);
     } catch (Exception e) {
```
vkkpng-g, line 45: Do we know the exact exception, or is there too many to consider?

capstone/src/test/java/com/google/sps/DetectShotsTest.java
```diff
@@ -0,0 +1,88 @@
...
+    detectShots.detect("gs://keyframe-images/download.png");
+  }
+
+  @Test (expected = Exception.class)
+  public void noFileWithPath() throws Exception {
+    detectShots.detect("gs://video-vigilance-videos/missing-video.mp4");
+  }
+
+  @Test
+  public void connectToAPI() throws Exception {
```
vkkpng-g, line 78: Add a test case for the nanosecond part of ur code

double startTime = segment.getStartTimeOffset().getSeconds()
                + segment.getStartTimeOffset().getNanos() / 1e9;
            double endTime = segment.getEndTimeOffset().getSeconds()

sharond106:
Since I am creating a fake shots list and just returning it when the function is called (with the mock when().thenReturn()), why is it necessary to test for empty list/ more than 1 shot in the list? I would just be confirming that the fake shot list did not change since I'm not currently testing how I handle the returned shots. Did I write my original test incorrectly?

capstone/src/main/java/com/google/sps/servlets/ShotsServlet.java
sharond106: I think there's too many to consider. But either way, I think all the exceptions would want to be handled the same way and let the user submit another file or use the backup way of getting shot times.

capstone/src/test/java/com/google/sps/DetectShotsTest.java
```diff
@@ -0,0 +1,88 @@
...
+import com.google.sps.data.DetectShots;
+import com.google.sps.data.Shot;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.List;
+import org.junit.Assert;
+import org.junit.Before;
```
cassmarcussen, line 26: Alphabetize junit imports

capstone/src/test/java/com/google/sps/DetectShotsTest.java
```diff
@@ -0,0 +1,88 @@
...
+    detectShots.detect("gs://video-vigilance-videos/missing-video.mp4");
+  }
+
+  @Test
+  public void connectToAPI() throws Exception {
+    ArrayList<Shot> shotslist = new ArrayList<Shot>();
+    Shot shot = new Shot(1, 4);
+    shotslist.add(shot);
+
+    when(mockDetectShots.detect(any(String.class))).thenReturn(shotslist);
```
cassmarcussen, line 83: Could you leave a comment about what this calls/does/tests, for some added documentation?

capstone/src/main/java/com/google/sps/data/DetectShots.java
sharond106: added

capstone/src/main/java/com/google/sps/data/DetectShots.java
sharond106: added

capstone/src/test/java/com/google/sps/DetectShotsTest.java
sharond106: how do i test this part of the code? it's past the point where the API permissions error will be thrown inside the method

capstone/src/test/java/com/google/sps/DetectShotsTest.java
```diff
@@ -0,0 +1,122 @@
...
+  private DetectShots detectShots;
+  private DetectShots mockDetectShots;
+
+  @Before
+  public void setup() {
+    detectShots = new DetectShots();
+    mockDetectShots = mock(DetectShots.class);
+  }
+
+  @Test (expected = Exception.class)
```
rwackes, line 44: Add a comment for each @Test, explaining what you are testing and what your expected result is. Something like: 
// A properly formatted GCS bucket path should start with "gs://". This path is incorrect because it starts with "gs:/".
// When an improperly formatted GCS bucket path is passed in, an Exception should be thrown.

capstone/src/test/java/com/google/sps/DetectShotsTest.java
```diff
@@ -0,0 +1,122 @@
...
+    when(mockDetectShots.detect(any(String.class))).thenReturn(shotslist);
+
+    ArrayList<Shot> shots = mockDetectShots.detect("gs://empty-bucket-for-tests");
+
+    String expected = "[{\"start_time\":1.0,\"end_time\":4.0}]";
+    Assert.assertEquals(expected, toJson(shots));
+  }
+
+  @Test
+  public void multipleShotsReturned() throws Exception {
```
rwackes, line 103: Does these tests using the mockDetectShots work or do they give you a Permission Denied Error?

sharond106: all my tests pass!

capstone/src/test/java/com/google/sps/DetectShotsTest.java
sharond106: added

capstone/src/test/java/com/google/sps/DetectShotsTest.java
```diff
@@ -0,0 +1,133 @@
...
+
+  // Set up mock DetectShots class before each test case
+  @Before
+  public void setup() {
+    detectShots = new DetectShots();
+    mockDetectShots = mock(DetectShots.class);
+  }
+  
+  // Url missing a slash
+  @Test (expected = Exception.class)
```
wdi-gg, line 46: Which lines of our code check Url formatting?

capstone/src/test/java/com/google/sps/DetectShotsTest.java
```diff
@@ -0,0 +1,133 @@
...
+  }
+
+  // Url leads to file that doesn't exist
+  @Test (expected = Exception.class)
+  public void noFileWithPath() throws Exception {
+    detectShots.detect("gs://video-vigilance-videos/missing-video.mp4");
+  }
+  
+  // Helper function that converts ArrayList of Shot objects to a json object
+  private String toJson(ArrayList<Shot> shots) {
```
wdi-gg, line 82: Move private helper function to the bottom.

capstone/src/test/java/com/google/sps/DetectShotsTest.java
```diff
@@ -0,0 +1,133 @@
...
+    detectShots.detect("gs:video-vigilance-videos");
+  }
+
+  // Url missing colon
+  @Test (expected = Exception.class)
+  public void incorrectBucketPathFormat3() throws Exception {
+    detectShots.detect("gs//video-vigilance-videos");
+  }
+ 
+  // Url leads to bucket that doesn't exist
```
wdi-gg, line 63: Will we be able to tell what are the actual exceptions? Just thinking they might through exceptions but not the expected types.

capstone/src/test/java/com/google/sps/DetectShotsTest.java
```diff
@@ -0,0 +1,133 @@
...
+  
+  // Helper function that converts ArrayList of Shot objects to a json object
+  private String toJson(ArrayList<Shot> shots) {
+    Gson gson = new Gson();
+    return gson.toJson(shots);
+  }
+
+  // Test when API returns no shots
+  @Test
+  public void noShotsReturned() throws Exception {
```
wdi-gg, line 89: Wait... Are we mocking our own class and function? Then we are not actually testing the code, right? Shouldn't we be mocking "client.annotateVideoAsync(request)"?

capstone/src/test/java/com/google/sps/DetectShotsTest.java
sharond106: I actually don't check for formatting. The detect() method in DetectShots.java will throw an Exception if the format is incorrect in any way or even if the format is correct, it may lead to an empty bucket or incorrect file type. It would be tricky for myself to explicitly check before passing it to this function. Are all the format checking tests unnecessary then?

capstone/src/test/java/com/google/sps/DetectShotsTest.java
sharond106: fixed!

capstone/src/test/java/com/google/sps/DetectShotsTest.java
sharond106: There are multiple exceptions that could occur while connecting to the API, so the detect() method just throws a generic Exception for any case.

capstone/src/test/java/com/google/sps/DetectShotsTest.java
sharond106: Yes this is what I was confused about in my comment above when I first started testing. I will work on this

capstone/src/test/java/com/google/sps/DetectShotsTest.java
vkkpng-g: I believe u tested so resolving

capstone/src/test/java/com/google/sps/DetectShotsTest.java
wdi-gg: Yes, we only need to test our own logic, otherwise there isn't a clear line of what to test.

capstone/src/test/java/com/google/sps/DetectShotsTest.java
```diff
@@ -0,0 +1,216 @@
...
+
+  // Url leads to file that doesn't exist
+  @Test (expected = Exception.class)
+  public void noFileWithPath() throws Exception {
+    detectShots.detect("gs://video-vigilance-videos/missing-video.mp4");
+  }
+
+  // Test when API returns no shots
+  @Test
+  public void noShotsReturned() throws Exception {
```
wdi-gg, line 114: I see we are only testing detect(), but let's also follow the test naming practice.

capstone/src/test/java/com/google/sps/DetectShotsTest.java
sharond106: Ok. I deleted all the tests that deal with formatting url

capstone/src/test/java/com/google/sps/DetectShotsTest.java
sharond106: fixed

PR #9, opened 2020-07-07
Setting shot times manually

sharond106:
I implemented a feature that allows the user to choose how often to capture frames. This is in the case that the Video Intelligence API runs into any errors/ no shots were detected by it. An alert box will pop up and ask the user to input a time in seconds or try to upload another video file. If the requested time interval is longer than the actual video, only the very last frame of the video will be captured.

Ex grabbing frames every 10 seconds: https://screenshot.googleplex.com/LT0HLWNDON0 

There is also a new button that captures a frame at the video's current paused time. The user can play the video, pause, and then capture to allow for more flexibility in what frames to capture.


capstone/src/main/webapp/js/extract-images-script.js
```diff
@@ -51,29 +54,55 @@ function getShots() {
...
-    document.getElementById("frames-list").appendChild(li);
-    return;
+    document.getElementById("frames-list").innerHTML = "No shots returned from Video Intelligence API.<br>";
+    
+    // Ask user for the time interval between frames to capture or Cancel
+    frameInterval = promptNumberInput();
+    if (isNaN(frameInterval)) {
+      return;
+    }
+    //If user did not Cancel and inputted a valid number of frames, call function to capture frames
```
rwackes, line 70: space after //

sharond106: oo good catch

capstone/src/main/webapp/js/extract-images-script.js
```diff
@@ -51,29 +54,55 @@ function getShots() {
...
   if (keyTimes.length == 0) {
-    const li = document.createElement("li");
-    li.innerHTML += "No shots to display<br>";
-    document.getElementById("frames-list").appendChild(li);
-    return;
+    document.getElementById("frames-list").innerHTML = "No shots returned from Video Intelligence API.<br>";
+    
+    // Ask user for the time interval between frames to capture or Cancel
+    frameInterval = promptNumberInput();
+    if (isNaN(frameInterval)) {
```
vkkpng-g, line 67: what does isNan do? I tried to look it up but couldn't really understand it.

capstone/src/main/webapp/js/extract-images-script.js
```diff
@@ -51,29 +54,55 @@ function getShots() {
 
 // Gets the first frame in the video by calling captureFrame
 function firstFrame() {
+  const path = URL.createObjectURL(document.querySelector("#video-file").files[0]);
+  
+  document.getElementById("frames-list").innerHTML = "";
+
   // If there are no shots to display, show error message
   if (keyTimes.length == 0) {
```
vkkpng-g, line 62: should we extract to separate method?

This section of code seems to be "if not keyTime, ask use for input"

Therefore method would called "promptUserInputForTime"/something like that

capstone/src/main/webapp/js/extract-images-script.js
```diff
@@ -51,29 +54,55 @@ function getShots() {
...
+    // If user did not Cancel and inputted a valid number of frames, call function to capture frames
+    document.getElementById("frames-list").innerHTML += "Capturing frames every " + frameInterval + " seconds.";
+    captureFrame(path, frameInterval);
   } 
-  // Otherwise, initialize variables
+  // If shots array is not empty, initialize variables and call function to capture frames
   else {
     keyTimesIndex = 0;
-    document.getElementById("frames-list").innerHTML = "";
+    frameInterval = -1;
```
vkkpng-g, line 77: Why do we set to -1?

capstone/src/main/webapp/js/extract-images-script.js
```diff
@@ -20,6 +20,9 @@ const keyTimes = [];
 // Current index of keyTimes
 var keyTimesIndex = 0;
 
+// Time interval between frames for manually setting shot times (-1 if not using this method)
+var frameInterval = -1;
```
vkkpng-g, line 24: It seems like this is only used in 1 method so there isn't a need to make it a global?

vkkpng-g, line 24: Can we add some comment when this global is being used. I see it used differently a couple of times. (like it serves for different purposes in diff methods)

capstone/src/main/webapp/js/extract-images-script.js
```diff
@@ -132,16 +165,33 @@ function displayFrame(img, secs, event) {
...
-    captureFrame(
-      URL.createObjectURL(document.querySelector("#video-file").files[0]),
-      keyTimes[keyTimesIndex]
-    );
-  };
+  // Check if there are more frames to capture, depending on which method of shot detection was used
+  if (frameInterval != -1 && (secs + frameInterval <= video.duration)) {
+    captureFrame(video.src, secs + frameInterval);
+  }
+  else if (++keyTimesIndex < keyTimes.length) {
```
vkkpng-g, line 174: What shot detection is this vs the if

capstone/src/main/webapp/js/extract-images-script.js
sharond106: isNaN is a javascript function that tests if the input "is Not A Number". So this part is just making sure the user inputted a valid number

capstone/src/main/webapp/js/extract-images-script.js
sharond106: That makes sense. I moved it to another method

capstone/src/main/webapp/js/extract-images-script.js
sharond106: It starts at -1 because whenever this variable is a valid frame interval time (positive number), then the displayFrame function will determine if there's more shots to capture differently. 

In the previous PR, at the end of the displayFrame function, I check if there's more shot times left in the keyTimes array. This PR introduces the new method of capturing shots by using a time interval, so the check for whether there are more shots to be captured needs to be different

capstone/src/main/webapp/js/extract-images-script.js
sharond106: Yes, I added more to the comment

capstone/src/main/webapp/js/extract-images-script.js
sharond106: I added comments in the code that hopefully clarifies this

capstone/src/main/webapp/js/extract-images-script.js
```diff
@@ -54,27 +57,60 @@ function getShots() {
...
     keyTimesIndex = 0;
+    frameInterval = -1;
     document.getElementById("frames-list").innerHTML = "";
-    path = URL.createObjectURL(document.querySelector("#video-file").files[0]);
+    captureFrame(path, keyTimes[keyTimesIndex]);
+  }
+}
+
+// Prompts the user for a time interval
+function promptUserForTime() {
```
wdi-gg, line 77: Logic looks good, and just a tip for readability.
This function doesn't only promptUserForTime, but also captureFrame. 
Instead of having a recursion call (checkForShots -> promptUserForTime -> captureFrame), it's generally more preferable to have main logic in a main function (e.g. checkForShots), and perform each task in a helper function (e.g. calling promptUserForTime, captureFrame one by one). In this way, readers will be able to grasp the main idea by reading only the main function.

capstone/src/main/webapp/js/extract-images-script.js
```diff
@@ -54,27 +57,60 @@ function getShots() {
...
+  path = URL.createObjectURL(document.querySelector("#video-file").files[0]);
+  
   if (keyTimes.length == 0 || !document.getElementById("video-file").value) {
     // If there are no shots to display or no file is selected, show error message
-    const li = document.createElement("li");
-    li.innerHTML += "No shots to display<br>";
-    document.getElementById("frames-list").appendChild(li);
-    return;
+    document.getElementById("frames-list").innerHTML = "No shots returned from Video Intelligence API.<br>";
+    promptUserForTime();
```
wdi-gg, line 65: Will user be able to go back to manual mode if they accidentally close the prompt?

capstone/src/main/webapp/js/extract-images-script.js
vkkpng-g: I think its a bit confusing that frameInterval is being used as a boolean as well as a value.

Let's do this for readibility

var getFramesByUserInput = false
var userInputFramIntervals = -1

capstone/src/main/webapp/js/extract-images-script.js
```diff
@@ -54,27 +57,60 @@ function getShots() {
...
+
+// Prompts the user for a time interval
+function promptUserForTime() {
+  frameInterval = promptNumberInput();
+  // If promptNumberInput() returned NaN, reset frameInterval to -1 and return
+  if (isNaN(frameInterval)) {
+    frameInterval = -1;
+    return;
   }
-  captureFrame(path, keyTimes[keyTimesIndex]);
```
vkkpng-g, line 70: getFramesByUserInput = true;

capstone/src/main/webapp/js/extract-images-script.js
```diff
@@ -129,13 +171,42 @@ function displayFrame(img, secs, event) {
...
+  // Check if there are more frames to capture, depending on which method of shot detection was used
+  // If frameInterval is not -1, this means the keyTimes array was empty and the user had to input a time interval
+  // To check if there are more frames to capure, see if going to the next frameInterval exceeds the video's end
+  // Ex. 
+  //    frameInterval = 5 s.
+  //    video.duration = 12 s.
+  //    secs = 10 s. (The last frame captured was at second 10)
+  //    
+  //    The next frame would be at 15 s., but since this is > 12 s., do not capture another frame
+  if (frameInterval != -1 && (secs + frameInterval <= video.duration)) {
```
vkkpng-g, line 185: add a var

var validNextFrame = secs + frameInterval <= video.duration
if (getFramesByUserInput && validNextFrame)

capstone/src/main/webapp/js/extract-images-script.js
```diff
@@ -129,13 +171,42 @@ function displayFrame(img, secs, event) {
...
+    captureFrame(video.src, secs + frameInterval);
+  }
+  // If frameInterval is -1, this means the keyTimes array was not empty and all times in the array should be captured
+  else if (++keyTimesIndex < keyTimes.length) {
```
vkkpng-g, line 189: Add a comment specifying why we are doing ++keyTimesIndex vs keyTimesIndex++.

capstone/src/main/webapp/js/extract-images-script.js
```diff
@@ -118,7 +157,10 @@ function captureFrame(path, secs) {
  * @param {event} event: Either a seeked event or an error event that called this function
  */
 function displayFrame(img, secs, event) {
```
vkkpng-g, line 159: Is it possible to not go into the if-else statement even though user goes into display frame?

If it is, should we create all these html aftewards?

keep a variable for the source and time
Do the if else --> capture the source and time in the vars
then do the html
then call capture frame

capstone/src/main/webapp/js/extract-images-script.js
sharond106: ok ya I agree

capstone/src/main/webapp/js/extract-images-script.js
```diff
@@ -54,27 +57,60 @@ function getShots() {
...
+  const message = "Error detecting shots in video. Check that your format is one of the following:" +
+                  "\n.MOV, .MPEG4, .MP4, .AVI, formats decodable by ffmpeg. \n\n" + 
+                  "Enter the time interval (in seconds) between image frames to analyze or " + 
+                  "click Cancel to submit another file.";
+  const defaultInput = 5;
+  var input = "";
+  // Reprompt user for input if input was not a number and did not Cancel prompt
+  do {
+    input = prompt(message, defaultInput);
+  } while (input != null && isNaN(input));
```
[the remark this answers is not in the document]
sharond106: added!

capstone/src/main/webapp/js/extract-images-script.js
sharond106: oh that makes sense! I moved the captureFrame call to checkForShots

capstone/src/main/webapp/js/extract-images-script.js
sharond106: In this version, they can hit the button that says 'show images' and this button will call checkForShots. So they will be able to redo it. In a later version, I think I will have to handle this differently since I get rid of this button and use an options menu instead. I'll make a note of this for the final version!

capstone/src/main/webapp/js/extract-images-script.js
sharond106: i'm a little confused. would I call captureFrame twice in displayFrame?

capstone/src/main/webapp/js/extract-images-script.js
wdi-gg: SG.

capstone/src/main/webapp/js/extract-images-script.js
vkkpng-g: ignore this comment. Did not have a good understanding of ur structure previously.

PR #10, opened 2020-07-07
Unit Testing for the Keyframe Image Upload Java Servlet

cassmarcussen:
### This PR includes the following tests:
- Tests for the POST and GET methods of the Keyframe Image Upload Servlet
- Tests for the POST method of the Keyframe Image Delete Servlet (i.e. the delete method of keyframe images)
- Tests for calling the DetectSafeSearchGcs' detectSafeSearchGcs() method, including tests for incorrect bucket url path formats
- A Mockito Controller which mocks http request and response parameters and calls the POST and GET methods for KeyframeImageUploadServlet and KeyframeImageDeleteServlet


capstone/src/test/java/com/google/sps/DetectSafeSearchGcsTest.java
```diff
@@ -0,0 +1,100 @@
...
+    mockDetectSafeSearch = mock(DetectSafeSearchGcs.class);
+  }
+
+  @Test (expected = Exception.class)
+  public void incorrectBucketPathFormat() throws Exception {
+    detectSafeSearch.detectSafeSearchGcs("gs:/video-vigilance-videos");
+  }
+
+  @Test (expected = Exception.class)
+  public void incorrectBucketPathFormat2() throws Exception {
```
vkkpng-g, line 59: in the test write why its incorrect. Same as below.

capstone/src/test/java/com/google/sps/DetectSafeSearchGcsTest.java
```diff
@@ -0,0 +1,100 @@
...
+    detectSafeSearch.detectSafeSearchGcs("gs://keyframe-images/download.png");
+  }
+
+  @Test (expected = Exception.class)
+  public void noFileWithPath() throws Exception {
+    detectSafeSearch.detectSafeSearchGcs("gs://video-vigilance-videos/missing-video.mp4");
+  }
+
+  @Test
+  public void connectToAPI() throws Exception {
```
vkkpng-g, line 84: Also add test for other cases like
1) error case
2) checking things are closed

capstone/src/test/java/com/google/sps/KeyframeImageMockitoController.java
```diff
@@ -0,0 +1,98 @@
...
+        request.addParameter("startTime", startTime);
+        request.addParameter("endTime", endTime);
+
+        try {
+            uploadServlet.doPost(request, response);
+        } catch (Exception e) {
+            System.err.println("exception: " + e);
+        }
+
+        System.out.println("POSTING...");
```
vkkpng-g, line 54: Removing print statement

capstone/src/test/java/com/google/sps/KeyframeImageMockitoController.java
```diff
@@ -0,0 +1,98 @@
...
+        MockHttpServletResponse response = new MockHttpServletResponse();
+
+        try {
+            uploadServlet.doGet(request, response);
+        } catch (Exception e) {
+            System.err.println("exception: " + e);
+        }
+
+        System.out.println("GET...");
+        return "get...";
```
vkkpng-g, line 75: In the test, you are testing status. Can we also test this return statement?

capstone/src/test/java/com/google/sps/KeyframeImageServletTest.java
```diff
@@ -0,0 +1,105 @@
...
+KeyframeImageServletTest tests the POST and GET requests of KeyframeImageUploadServlet, which correspondingly 
+post keyframe images and their corresponding information to DataStore and Google Cloud Bucket and retrieve the keyframe 
+images and their corresponding information from DataStore and GCB. KeyframeImageServletTest also tests the POST request 
+of KeyframeImageDeleteServlet, which deletes all keyframe images and their corresponding information from DataStore and 
+Google Cloud Bucket. This delete serves as a refresh to be called at the start of the user flow.
+*/
+public class KeyframeImageServletTest {
+
+    private MockMvc mockMvc;
+
```
vkkpng-g, line 49: Test failure case for each method + also the string that you returned for each method.

The reason is because we don't have 100% code testing.

capstone/src/test/java/com/google/sps/DetectSafeSearchGcsTest.java
```diff
@@ -0,0 +1,100 @@
...
+package com.google.sps.servlets;
+
+import com.google.sps.servlets.DetectSafeSearchGcs;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
```
sharond106, line 24: Are you using arrayList, arrays, collections, and list in this file?

capstone/src/test/java/com/google/sps/KeyframeImageServletTest.java
```diff
@@ -0,0 +1,105 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.sps.servlets;
+
+import java.util.ArrayList;
```
sharond106, line 17: Double check if you need all these imports too

capstone/src/test/java/com/google/sps/DetectSafeSearchGcsTest.java
cassmarcussen: Hi! I was wondering if you could elaborate on this comment. I'm not sure what testing "error case" means in this context.

capstone/src/test/java/com/google/sps/KeyframeImageServletTest.java
cassmarcussen: Hi! I was wondering if you could elaborate on this comment. Are you saying that the reason we want to test failure cases for each method and the returned string is because we want 100% code testing?

capstone/src/test/java/com/google/sps/DetectSafeSearchGcsTest.java
```diff
@@ -0,0 +1,95 @@
...
+  }
+
+  @Test (expected = Exception.class)
+  public void incorrectBucketPathFormat2() throws Exception {
+    // This test is incorrect because the 'gs' is followed by ':' instead of '://' in the url
+    detectSafeSearch.detectSafeSearchGcs("gs:video-vigilance-videos");
+  }
+
+  @Test (expected = Exception.class)
+  public void incorrectBucketPathFormat3() throws Exception {
```
vkkpng-g, line 61: Sorry meant like

incorrectBucketPathMissingSemiColon

capstone/src/test/java/com/google/sps/DetectSafeSearchGcsTest.java
vkkpng-g: incorrectBucketPathMissingDoubleSlash

capstone/src/test/java/com/google/sps/DetectSafeSearchGcsTest.java
```diff
@@ -0,0 +1,95 @@
...
+  private DetectSafeSearchGcs mockDetectSafeSearch;
+
+  @Before
+  public void setup() {
+    detectSafeSearch = new DetectSafeSearchGcs();
+    mockDetectSafeSearch = mock(DetectSafeSearchGcs.class);
+  }
+
+  @Test (expected = Exception.class)
+  public void incorrectBucketPathFormat() throws Exception {
```
vkkpng-g, line 49: incorrectBucketPathMissingOneSlash

The test names should correlate to what you are testing

capstone/src/test/java/com/google/sps/DetectSafeSearchGcsTest.java
vkkpng-g: For error case:

So i believe in your method you have a code flow where the AnnotateImageResponse has errors
so for each value in the HashMap, you return "UNKNOWN".

vkkpng-g: For closing, just making sure the client is closed.

capstone/src/test/java/com/google/sps/KeyframeImageDataStoreTest.java
```diff
@@ -0,0 +1,238 @@
...
+  // Testing local datastore service with no entities
+  @Test
+  public void noEntities() {
+    DatastoreService dataService = DatastoreServiceFactory.getDatastoreService();
+    Query query = new Query("KeyframeImages_Video_TestList");
+    PreparedQuery results = dataService.prepare(query);
+    Assert.assertEquals(0, results.countEntities(withLimit(10)));
+  }
+
+  // Posting null url
```
vkkpng-g, line 65: Remove the commented code?

capstone/src/test/java/com/google/sps/KeyframeImageDataStoreTest.java
```diff
@@ -0,0 +1,238 @@
...
+    String startTime = "0:00";
+    String endTime = "0:15";
+    keyframeImageUpload.createAndPostEntity(testUrl, timestamp, startTime, endTime, "KeyframeImages_Video_TestList");
+
+    Assert.assertEquals(1, dataService.prepare(new Query("KeyframeImages_Video_TestList")).countEntities(withLimit(10)));
+
+    // The asSingleEntity() function retrieves the one and only result for the Query
+    Entity queryResult = dataService.prepare(new Query("KeyframeImages_Video_TestList")).asSingleEntity();
+    // Check url property 
+    Assert.assertEquals(testUrl, queryResult.getProperty("url"));
```
vkkpng-g, line 115: why are we checking the url property?

capstone/src/test/java/com/google/sps/KeyframeImageDataStoreTest.java
cassmarcussen: DataStore testing is work-in-progress.

capstone/src/test/java/com/google/sps/KeyframeImageDataStoreTest.java
cassmarcussen: DataStore testing is work-in-progress. I still need to modify this.

capstone/src/test/java/com/google/sps/DetectSafeSearchGcsTest.java
```diff
@@ -0,0 +1,95 @@
...
+  private DetectSafeSearchGcs detectSafeSearch;
+  private DetectSafeSearchGcs mockDetectSafeSearch;
+
+  @Before
+  public void setup() {
+    detectSafeSearch = new DetectSafeSearchGcs();
+    mockDetectSafeSearch = mock(DetectSafeSearchGcs.class);
+  }
+
+  @Test (expected = Exception.class)
```
wdi-gg, line 48: Ah I think many of the comments in https://github.com/cassmarcussen/video-vigilance/pull/7 and https://github.com/cassmarcussen/video-vigilance/pull/21 can also be applied to this PR. Could you take a look?
e.g.
1. Test methods are normally named as testFunc_testCase, in our case they should be testDetectSafeSearchGcs_incorrectBucketPathOneSlash(), etc.
2. Which lines of our code check Url formatting?
3. Will we be able to tell what are the actual exceptions? Just thinking they might through exceptions but not the expected types.
...

capstone/src/test/java/com/google/sps/DetectSafeSearchGcsTest.java
```diff
@@ -0,0 +1,95 @@
...
+  public void noFileWithPath() throws Exception {
+    // This test is incorrect because the file missing-image.jpg does not exist in the keyframe-images bucket
+    detectSafeSearch.detectSafeSearchGcs("gs://video-vigilance-videos/missing-image.jpg");
+  }
+
+  /*@Test
+  public void connectToAPI() throws Exception {
```
wdi-gg, line 79: I see "Returns permission denied error". Will we be able to run the tests when deployed?

capstone/src/test/java/com/google/sps/DetectSafeSearchGcsTest.java
vkkpng-g: Any updates with new tests for DetectSafeSearchGcs?

capstone/src/test/java/com/google/sps/KeyframeImageDataStoreTest.java
```diff
@@ -0,0 +1,234 @@
...
+    helper.setUp();
+  }
+  @After
+  public void tearDown() {
+    helper.tearDown();
+  }
+
+  // Testing local datastore service with no entities
+  @Test
+  public void noEntities() {
```
vkkpng-g, line 58: Is this test testing any of the code u wrote?

capstone/src/test/java/com/google/sps/KeyframeImageDataStoreTest.java
```diff
@@ -0,0 +1,234 @@
...
+    Query query = new Query("KeyframeImages_Video_TestList");
+    PreparedQuery results = dataService.prepare(query);
+    Assert.assertEquals(0, results.countEntities(withLimit(10)));
+  }
+
+  // Posting 1 entity 
+  @Test
+  public void addOneEntityWithProperty() {
+    DatastoreService dataService = DatastoreServiceFactory.getDatastoreService();
+    Assert.assertEquals(0, dataService.prepare(new Query("KeyframeImages_Video_TestList")).countEntities(withLimit(10)));
```
vkkpng-g, line 69: why do we have .countEntities(withLimit(10)). I'm not sure what this L69 is testing.

capstone/src/test/java/com/google/sps/KeyframeImageDataStoreTest.java
```diff
@@ -0,0 +1,234 @@
...
+
+    KeyframeImageUploadServlet keyframeImageUpload = new KeyframeImageUploadServlet();
+    String testUrl = "fake.url";
+    String timestamp = "0:10";
+    String startTime = "0:00";
+    String endTime = "0:15";
+    keyframeImageUpload.createAndPostEntity(testUrl, timestamp, startTime, endTime, "KeyframeImages_Video_TestList");
+
+    Assert.assertEquals(1, dataService.prepare(new Query("KeyframeImages_Video_TestList")).countEntities(withLimit(10)));
```
vkkpng-g, line 78: What are we testing in this line?

vkkpng-g: It seems like we are checking something is in the datastore? Why do we limit by 10?

capstone/src/test/java/com/google/sps/KeyframeImageDataStoreTest.java
```diff
@@ -0,0 +1,234 @@
...
+    Assert.assertEquals(endTime, queryResult.getProperty("endTime"));
+  }
+
+  // Posting multiple entities
+  @Test
+  public void addMultipleEntities() {
+    DatastoreService dataService = DatastoreServiceFactory.getDatastoreService();
+    Assert.assertEquals(0, dataService.prepare(new Query("KeyframeImages_Video_TestList")).countEntities(withLimit(10)));
+
+    KeyframeImageUploadServlet keyframeImageUpload = new KeyframeImageUploadServlet();
```
vkkpng-g, line 95: I prefer L96 of create the post request rather than L71-75

capstone/src/test/java/com/google/sps/KeyframeImageDataStoreTest.java
```diff
@@ -0,0 +1,234 @@
...
+    keyframeImageUpload.createAndPostEntity("fake.url.1", "0:10", "0:00", "0:15", "KeyframeImages_Video_TestList");
+    keyframeImageUpload.createAndPostEntity("fake.url.2", "0:25", "0:20", "0:30", "KeyframeImages_Video_TestList");
+    keyframeImageUpload.createAndPostEntity("fake.url.3", "0:40", "0:35", "0:45", "KeyframeImages_Video_TestList");
+
+    Query query = new Query("KeyframeImages_Video_TestList");
+    PreparedQuery results = dataService.prepare(query);
+
+    Assert.assertEquals(3, results.countEntities(withLimit(10)));
```
vkkpng-g, line 103: Same why limit 10?

capstone/src/test/java/com/google/sps/KeyframeImageDataStoreTest.java
```diff
@@ -0,0 +1,234 @@
...
+
+  // Getting from datastore with multiple entities
+  @Test
+  public void getMultipleImagesAndQuerySortByTimestamp() {
+    DatastoreService dataService = DatastoreServiceFactory.getDatastoreService();
+    Assert.assertEquals(0, dataService.prepare(new Query("KeyframeImages_Video_TestList")).countEntities(withLimit(10)));
+
+    // Add entity to datastore
+    
+    Entity entity1 = new Entity("KeyframeImages_Video_TestList");
```
vkkpng-g, line 165: Lets create a helper method to create these things from L165-174

capstone/src/test/java/com/google/sps/KeyframeImageDataStoreTest.java
```diff
@@ -0,0 +1,234 @@
...
+    String timestamp2 = "0:10";
+    String startTime2 = "0:00";
+    String endTime2 = "0:15";
+    entity2.setProperty("url", testUrl2);
+    entity2.setProperty("timestamp", timestamp2);
+    entity2.setProperty("startTime", startTime2);
+    entity2.setProperty("endTime", endTime2);
+    dataService.put(entity2);
+
+    Entity entity3 = new Entity("KeyframeImages_Video_TestList");
```
vkkpng-g, line 187: lets name ur entity better. They seemed to be different y the timestamp so perhaps a naming that signifies that behavior?

capstone/src/test/java/com/google/sps/KeyframeImageDataStoreTest.java
```diff
@@ -0,0 +1,234 @@
...
+    entity3.setProperty("startTime", startTime3);
+    entity3.setProperty("endTime", endTime3);
+    dataService.put(entity3);
+
+
+    List<KeyframeImage> listOfOneKeyframeImage = new ArrayList<>();
+    // In the GET method for DataStore, we add "gs:/" to the front of the URL of the Keyframe Image so it can be properly 
+    // served on the page and for the Vision API, so in testing we must add it to the beginning of our test URL
+    // Add these in the order they should appear for timestamp ascending order sort of query
+    KeyframeImage testKeyframeImageFirstTimestamp = new KeyframeImage("gs:/" + testUrl2, timestamp2, startTime2, endTime2);
```
vkkpng-g, line 203: since it seems like u need all these variables, lets also define them in one line

string firstUrl, SecondUrl,ThirdUrl = "x", "x", "x';
String firstStart, Second, ext = "x', "x', "x'

capstone/src/test/java/com/google/sps/KeyframeImageDataStoreTest.java
```diff
@@ -0,0 +1,234 @@
...
+    List<KeyframeImage> listOfOneKeyframeImage = new ArrayList<>();
+    // In the GET method for DataStore, we add "gs:/" to the front of the URL of the Keyframe Image so it can be properly 
+    // served on the page and for the Vision API, so in testing we must add it to the beginning of our test URL
+    KeyframeImage testKeyframeImage = new KeyframeImage("gs:/" + testUrl, timestamp, startTime, endTime);
+    listOfOneKeyframeImage.add(testKeyframeImage);
+
+    KeyframeImageUploadServlet keyframeImageUpload = new KeyframeImageUploadServlet();
+    List<KeyframeImage> listOfKeyframeImages = keyframeImageUpload.getKeyframeImagesFromDataStore("KeyframeImages_Video_TestList");
+    Assert.assertEquals(listOfOneKeyframeImage.size(), listOfKeyframeImages.size());
+    Assert.assertEquals(1, listOfOneKeyframeImage.size());
```
vkkpng-g, line 146: Whats the difference between listOfOneKeyFramImage vs listOfKeyframeImages

Why do we have to test both of them?

capstone/src/test/java/com/google/sps/KeyframeImageDataStoreTest.java
```diff
@@ -0,0 +1,234 @@
...
+
+    KeyframeImageUploadServlet keyframeImageUpload = new KeyframeImageUploadServlet();
+    List<KeyframeImage> listOfKeyframeImages = keyframeImageUpload.getKeyframeImagesFromDataStore("KeyframeImages_Video_TestList");
+    Assert.assertEquals(listOfOneKeyframeImage.size(), listOfKeyframeImages.size());
+    Assert.assertEquals(3, listOfOneKeyframeImage.size());
+    Assert.assertEquals(3, listOfKeyframeImages.size());
+
+    // The addresses of the keyframe images in the lists will be different, so we need to test each of the properties of the Keyframe Images
+    // This tests the sorting by timestamp in the Query
+    Assert.assertEquals(listOfOneKeyframeImage.get(1).getUrl(), listOfKeyframeImages.get(1).getUrl());
```
vkkpng-g, line 218: L218-231, lets create a helper method to test all those

Call XXX(listOfOneKeyframeImage.get(0), listOfKeyframeImages.get(0))
Call XXX(listOfOneKeyframeImage.get(1), listOfKeyframeImages.get(1))
Call XXX(listOfOneKeyframeImage.get(2), listOfKeyframeImages.get(2))

Test XXX method 
parameter: 1 listOfOneKeyframeImage, 1  listOfKeyframeImages
in the method: do the test from L223-226

capstone/src/test/java/com/google/sps/KeyframeImageServletTest.java
vkkpng-g: Yes we want to try to get 100% lines coverage/ not only lines but the different combinations that can happen

Just say you have 

if x
else y
if a
else b

Test: xa, xb, ya, yb (4 different possibilities)

capstone/src/test/java/com/google/sps/DetectSafeSearchGcsTest.java
cassmarcussen: Fixed

capstone/src/test/java/com/google/sps/KeyframeImageMockitoController.java
cassmarcussen: Fixed

capstone/src/test/java/com/google/sps/KeyframeImageMockitoController.java
cassmarcussen: Fixed

capstone/src/test/java/com/google/sps/DetectSafeSearchGcsTest.java
cassmarcussen: Fixed

capstone/src/test/java/com/google/sps/KeyframeImageServletTest.java
cassmarcussen: Fixed

capstone/src/test/java/com/google/sps/DetectSafeSearchGcsTest.java
cassmarcussen: Fixed

capstone/src/test/java/com/google/sps/DetectSafeSearchGcsTest.java
cassmarcussen: Fixed

capstone/src/test/java/com/google/sps/KeyframeImageDataStoreTest.java
```diff
@@ -0,0 +1,238 @@
+// Copyright 2019 Google LLC
```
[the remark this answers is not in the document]
cassmarcussen: Added a sorting test (In the new test, when we return multiple keyframe images, we expect them in a sorted order)

capstone/src/test/java/com/google/sps/DetectSafeSearchGcsTest.java
cassmarcussen: I think that 'mvn test' is something that is only run locally, in which case we will always get a "Returns permission denied error" because of the APIs. Is there a way to run the tests when deployed?

capstone/src/test/java/com/google/sps/KeyframeImageDataStoreTest.java
cassmarcussen: It is testing the edge case that we have a datastore list with no keyframe images/entities, which is relevant to my code/design.

capstone/src/test/java/com/google/sps/KeyframeImageDataStoreTest.java
cassmarcussen: For L71-75, I broke up each property of the entity because on L83-86 I am testing that each property is what we expect it to be. I was thinking that this would be more efficient and extensible, since if I change a value I will only have to change it in one place instead of two. L96 I am not checking the values, so I kept it all on one line. With this in mind should I convert L71-75 to the format of L96?

capstone/src/test/java/com/google/sps/KeyframeImageDataStoreTest.java
cassmarcussen: Fixed

capstone/src/test/java/com/google/sps/KeyframeImageDataStoreTest.java
cassmarcussen: No longer named, because created in the helper method.

capstone/src/test/java/com/google/sps/KeyframeImageDataStoreTest.java
cassmarcussen: Fixed

capstone/src/test/java/com/google/sps/KeyframeImageDataStoreTest.java
```diff
@@ -0,0 +1,234 @@
...
+    KeyframeImage testKeyframeImageSecondTimestamp = new KeyframeImage("gs:/" + testUrl1, timestamp1, startTime1, endTime1);
+    KeyframeImage testKeyframeImageThirdTimestamp = new KeyframeImage("gs:/" + testUrl3, timestamp3, startTime3, endTime3);
+    listOfOneKeyframeImage.add(testKeyframeImageFirstTimestamp);
+    listOfOneKeyframeImage.add(testKeyframeImageSecondTimestamp);
+    listOfOneKeyframeImage.add(testKeyframeImageThirdTimestamp);
+
+    KeyframeImageUploadServlet keyframeImageUpload = new KeyframeImageUploadServlet();
+    List<KeyframeImage> listOfKeyframeImages = keyframeImageUpload.getKeyframeImagesFromDataStore("KeyframeImages_Video_TestList");
+    Assert.assertEquals(listOfOneKeyframeImage.size(), listOfKeyframeImages.size());
+    Assert.assertEquals(3, listOfOneKeyframeImage.size());
```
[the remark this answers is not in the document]
cassmarcussen: It is a bit repeated in logic, but I wanted to make sure that both of the lists are of size 3 before comparing if they are the same size. For example, if the first of the lists is the wrong size but the second is the correct size, failing on these lines will give more specific error messages and tell us which list is the wrong size. 

But, L212 would fail before getting to these in this case, so maybe I can move L213-214 to before L212 so these fail first?

capstone/src/test/java/com/google/sps/KeyframeImageDataStoreTest.java
cassmarcussen: Renaming listOfOneKeyframeImages to listOfThreeKeyframeImages. Renaming listOfKeyframeImages to listOfKeyframeImagesFromDataStore.

listOfThreeKeyframeImages is a hard-coded list of three keyframe images. listOfKeyframeImagesFromDataStore is the list of keyframe images returned from datastore. I want to make sure that the values returned from datastore are what we expect, and in the correct order, which is why I hard-code listOfThreeKeyframeImages.

I'll leave a comment in the code about this.

capstone/src/test/java/com/google/sps/KeyframeImageDataStoreTest.java
```diff
@@ -0,0 +1,234 @@
...
+    DatastoreService dataService = DatastoreServiceFactory.getDatastoreService();
+    Assert.assertEquals(0, dataService.prepare(new Query("KeyframeImages_Video_TestList")).countEntities(withLimit(10)));
+
+    // Add entity to datastore
+    Entity entity = new Entity("KeyframeImages_Video_TestList");
+    String testUrl = "fake.url";
+    String timestamp = "0:10";
+    String startTime = "0:00";
+    String endTime = "0:15";
+    entity.setProperty("url", testUrl);
```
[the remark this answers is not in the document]
cassmarcussen: Fixed

capstone/src/test/java/com/google/sps/KeyframeImageDataStoreTest.java
cassmarcussen: Added

capstone/src/test/java/com/google/sps/DetectSafeSearchGcsTest.java
cassmarcussen: Sorry, I must have missed this. I'll address them today.

capstone/src/test/java/com/google/sps/KeyframeImageDataStoreTest.java
cassmarcussen: Removed withLimit. L69 is just testing that the DataStore list is empty before we add to it.

capstone/src/test/java/com/google/sps/KeyframeImageDataStoreTest.java
cassmarcussen: Removed withLimit(10), since it is not needed. We are testing that exactly 1 entity has been added to our DataStore list.

capstone/src/test/java/com/google/sps/KeyframeImageDataStoreTest.java
cassmarcussen: Removed

capstone/src/test/java/com/google/sps/KeyframeImageDataStoreTest.java
```diff
@@ -0,0 +1,214 @@
...
+
+    KeyframeImageUploadServlet keyframeImageUpload = new KeyframeImageUploadServlet();
+    String testUrl = "fake.url";
+    String timestamp = "0:10";
+    String startTime = "0:00";
+    String endTime = "0:15";
+    keyframeImageUpload.createAndPostEntity(testUrl, timestamp, startTime, endTime, "KeyframeImages_Video_TestList");
+
+    // Testing that exactly 1 entity has been added to our DataStore list
+    Assert.assertEquals(1, dataService.prepare(new Query("KeyframeImages_Video_TestList")).countEntities());
```
sharond106, line 80: Do all the tests here pass? If so , I'm wondering how your local dataService variable is getting the KeyframeImages_Video_TestList posted to it? In createAndPostEntity, it's initializing another datastore service variable, so it seems like there are 2 different ones being used here.

capstone/src/test/java/com/google/sps/KeyframeImageDataStoreTest.java
```diff
@@ -0,0 +1,214 @@
...
+    // The addresses of the keyframe images in the lists will be different, so we need to test each of the properties of the Keyframe Images
+    // This tests the sorting by timestamp in the Query
+    assertValuesEqual(listOfThreeKeyframeImages.get(1), listOfKeyframeImagesFromDataStore.get(1));
+    assertValuesEqual(listOfThreeKeyframeImages.get(0), listOfKeyframeImagesFromDataStore.get(0));
+    assertValuesEqual(listOfThreeKeyframeImages.get(2), listOfKeyframeImagesFromDataStore.get(2));
+
+  }
+
+  // Helper method to add an entity to DataStore
+  private void addEntityToDatastore(String url, String timestamp, String startTime, String endTime, DatastoreService dataService) {
```
sharond106, line 197: Maybe you can also add tests that use your servlet's get and post helper methods together in addition to the tests you have so far that test them separately

capstone/src/test/java/com/google/sps/KeyframeImageDataStoreTest.java
cassmarcussen: Yes the tests pass here. createAndPostEntity takes in a datastore list name in the parameters (the last parameter), and posts an entity to this datastore list. Line 80 here queries from the same DataStore list.

Where do you see 2 different list names being used?

capstone/src/test/java/com/google/sps/DetectSafeSearchGcsTest.java
```diff
@@ -0,0 +1,115 @@
...
+    when(mockDetectSafeSearch.detectSafeSearchGcs(anyString())).thenReturn(mockSafeSearchResults);
+
+    // NOTE: Returns permission denied error
+    HashMap<String, String> safeSearchResults = mockDetectSafeSearch.detectSafeSearchGcs("gs://keyframe-images-to-effect/AAANsUnmvLkSJZEVnYAh6DNG6O13zzRusbFKKRTwjdDj81ikKqNbo7wwYIvwYQUJd1bnQCW0XdNRjf82G21nk7yBGfqObtMJgw.R2GN-ZINyUODcEv1");
+    
+    Assert.assertEquals(mockSafeSearchResults, safeSearchResults);
+  }
+
+  public void connectToAPIErrorCase() throws Exception {
+    // May need to adjust test values if bucket gets deleted...
```
sharond106, line 97: I rewrote my tests for connecting to the Video Intelligence API for getting shot times in #7 (after you already approved so you didn't see them) and you may be able to use the same way I re structured my code and tested it

capstone/src/test/java/com/google/sps/KeyframeImageDataStoreTest.java
cassmarcussen: Sorry, could you clarify what you mean? In these tests I am using the helper methods for get and post.

capstone/src/test/java/com/google/sps/KeyframeImageDataStoreTest.java
sharond106: Oh ok. I meant that inside createAndPostEntity, line 124, i thought it was using another datastore and putting the new entity in that one.

capstone/src/test/java/com/google/sps/DetectSafeSearchGcsTest.java
cassmarcussen: Thanks, I'll take a look!

capstone/src/test/java/com/google/sps/KeyframeImageDataStoreTest.java
sharond106: I meant that you can use the servlet's createAndPost() to post an image and getKeyframeImagesFromDataStore() to get the image together instead of testing these methods in separate unit tests

capstone/src/test/java/com/google/sps/DetectSafeSearchGcsTest.java
cassmarcussen: Hi Sharon! I've tried to use your restructured method of testing out your API. Unfortunately, the Vision API gives me a PERMISSION_DENIED error, so I don't think this will work for my tests.

capstone/src/test/java/com/google/sps/KeyframeImageDataStoreTest.java
cassmarcussen: Ok, I can add that! I think for unit testing it's most typical for tests to just test out one function / the smallest unit at a time, which is why I didn't do this. But it would still be a helpful test!

cassmarcussen: Added

capstone/src/test/java/com/google/sps/KeyframeImageDataStoreTest.java
vkkpng-g: No worries, thanks for clearing it up. No need to change.

capstone/src/test/java/com/google/sps/KeyframeImageDataStoreTest.java
```diff
@@ -0,0 +1,234 @@
...
+  @Test
+  public void getListWithNoImages() {
+    DatastoreService dataService = DatastoreServiceFactory.getDatastoreService();
+    Assert.assertEquals(0, dataService.prepare(new Query("KeyframeImages_Video_TestList")).countEntities());
+
+    List<KeyframeImage> emptyListOfKeyframeImages = new ArrayList<>();
+
+    KeyframeImageUploadServlet keyframeImageUpload = new KeyframeImageUploadServlet();
+    List<KeyframeImage> listOfKeyframeImagesFromDataStore = keyframeImageUpload.getKeyframeImagesFromDataStore("KeyframeImages_Video_TestList");
+    Assert.assertEquals(emptyListOfKeyframeImages, listOfKeyframeImagesFromDataStore);
```
vkkpng-g, line 138: Instead of asserting equal..can u test that listOfKeyframeImagesFromDataStore is empty()?

capstone/src/test/java/com/google/sps/KeyframeImageDataStoreTest.java
vkkpng-g: +1 yes that is typical of unit test, but since we are not enforcing e2e test, adding them to these test is also good.

capstone/src/test/java/com/google/sps/KeyframeImageDataStoreTest.java
vkkpng-g: Update..it seems we are still checking the url property?

capstone/src/test/java/com/google/sps/KeyframeImageServletTest.java
vkkpng-g: I think u tested the output string.


Can we add tests that will trigger the IOException? The reason is because your code handles the exceptions but we don't ever tests that.

capstone/src/test/java/com/google/sps/KeyframeImageDataStoreTest.java
vkkpng-g: Hey Cassandra, so basically no need to test code that we didn't write. The reason is we should assume that they are following their own API (and they themselvees tested their code).

capstone/src/test/java/com/google/sps/KeyframeImageDataStoreTest.java
```diff
@@ -0,0 +1,234 @@
...
+    Entity entity = new Entity("KeyframeImages_Video_TestList");
+    entity.setProperty("url", url);
+    entity.setProperty("timestamp", timestamp);
+    entity.setProperty("startTime", startTime);
+    entity.setProperty("endTime", endTime);
+    dataService.put(entity); 
+  }
+
+  // Helper method which asserts two KeyframeImages to be equal by checking each of their fields
+  private void assertValuesEqual(KeyframeImage firstImage, KeyframeImage secondImage) {
```
vkkpng-g, line 227: Just a nit

the variables should probably be "expected", "action" something along those lines. In this helper method, we're not sure which one is coming from the test output.

capstone/src/test/java/com/google/sps/KeyframeImageDataStoreTest.java
```diff
@@ -0,0 +1,234 @@
...
+    KeyframeImage testKeyframeImageSecondTimestamp = new KeyframeImage("gs:/" + testUrl1, timestamp1, startTime1, endTime1);
+    KeyframeImage testKeyframeImageThirdTimestamp = new KeyframeImage("gs:/" + testUrl3, timestamp3, startTime3, endTime3);
+    listOfThreeKeyframeImages.add(testKeyframeImageFirstTimestamp);
+    listOfThreeKeyframeImages.add(testKeyframeImageSecondTimestamp);
+    listOfThreeKeyframeImages.add(testKeyframeImageThirdTimestamp);
+
+    KeyframeImageUploadServlet keyframeImageUpload = new KeyframeImageUploadServlet();
+    List<KeyframeImage> listOfKeyframeImagesFromDataStore = keyframeImageUpload.getKeyframeImagesFromDataStore("KeyframeImages_Video_TestList");
+    Assert.assertEquals(3, listOfThreeKeyframeImages.size());
+    Assert.assertEquals(3, listOfKeyframeImagesFromDataStore.size());
```
vkkpng-g, line 205: ?: Why do we test that both are equal to 3. The next line below already checks that both are equal.

Instead what we should do is
Assert the output from the call is 3
Assert the 2 objects are the same size

capstone/src/test/java/com/google/sps/KeyframeImageDataStoreTest.java
```diff
@@ -0,0 +1,234 @@
...
+    helper.setUp();
+  }
+  @After
+  public void tearDown() {
+    helper.tearDown();
+  }
+
+  // Testing the edge case that we have a datastore list with no keyframe images/entities
+  @Test
+  public void noEntities() {
```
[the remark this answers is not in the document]
cassmarcussen: Fixed

capstone/src/test/java/com/google/sps/KeyframeImageDataStoreTest.java
```diff
@@ -0,0 +1,234 @@
+// Copyright 2019 Google LLC
```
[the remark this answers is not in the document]
cassmarcussen: Fixed

capstone/src/test/java/com/google/sps/KeyframeImageDataStoreTest.java
cassmarcussen: Fixed

capstone/src/test/java/com/google/sps/KeyframeImageDataStoreTest.java
cassmarcussen: Fixed

capstone/src/test/java/com/google/sps/KeyframeImageDataStoreTest.java
cassmarcussen: Fixed

capstone/src/test/java/com/google/sps/KeyframeImageDataStoreTest.java
cassmarcussen: We are checking the following properties: 
Assert.assertEquals(testUrl, queryResult.getProperty("url"));
Assert.assertEquals(timestamp, queryResult.getProperty("timestamp"));
Assert.assertEquals(startTime, queryResult.getProperty("startTime"));
Assert.assertEquals(endTime, queryResult.getProperty("endTime"));

Once I merge this to a branch that doesn't have startTime and endTime I'll remove the checks for start and end time. I'm checking all the properties stored in DataStore to make sure they are all as we expect.

capstone/src/test/java/com/google/sps/DetectSafeSearchGcsTest.java
cassmarcussen: Hi! Sharon had also suggested I look at #7. I tried using the changes she had made, but unfortunately the Vision API still gives a permission denied error with the sort of changes as in #7.

capstone/src/test/java/com/google/sps/KeyframeImageDataStoreTest.java
cassmarcussen: Hi! I removed the test.

capstone/src/test/java/com/google/sps/KeyframeImageDataStoreTest.java
```diff
@@ -0,0 +1,234 @@
...
+    DatastoreService dataService = DatastoreServiceFactory.getDatastoreService();
+    Query query = new Query("KeyframeImages_Video_TestList");
+    PreparedQuery results = dataService.prepare(query);
+    Assert.assertEquals(0, results.countEntities());
+  }
+
+  // Posting 1 entity 
+  @Test
+  public void addOneEntityWithProperty() {
+    DatastoreService dataService = DatastoreServiceFactory.getDatastoreService();
```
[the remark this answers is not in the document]
cassmarcussen: Fixed, and looked at 21

capstone/src/test/java/com/google/sps/DetectSafeSearchGcsTest.java
```diff
@@ -0,0 +1,110 @@
...
+
+    public HashMap<String, String> detectSafeSearchGcs(String gcsPath) {
+      return new HashMap<String, String>();
+    }
+
+  }
+
+  @Before
+  public void setup() throws Exception {
+    detectSafeSearch = new DetectSafeSearchGcs();
```
vkkpng-g, line 58: Are we using detectSafeSearch?

capstone/src/test/java/com/google/sps/DetectSafeSearchGcsTest.java
```diff
@@ -0,0 +1,110 @@
...
+  @Test
+  public void testDetectSafeSearchGcs_noFileWithPath() throws Exception {
+    // This test is incorrect because the file missing-image.jpg does not exist in the keyframe-images bucket
+    mockDetectSafeSearch.detectSafeSearchGcs("gs://video-vigilance-videos/missing-image.jpg");
+  }
+
+
+  @Test
+  public void connectToAPI() throws Exception {
+    HashMap<String, String> safeSearchResults = mockDetectSafeSearch.detectSafeSearchGcs("gs://keyframe-images-to-effect/AAANsUnmvLkSJZEVnYAh6DNG6O13zzRusbFKKRTwjdDj81ikKqNbo7wwYIvwYQUJd1bnQCW0XdNRjf82G21nk7yBGfqObtMJgw.R2GN-ZINyUODcEv1");
```
vkkpng-g, line 105: This seems right..

capstone/src/test/java/com/google/sps/DetectSafeSearchGcsTest.java
```diff
@@ -0,0 +1,110 @@
...
+  @Test
+  public void testDetectSafeSearchGcs_nonexistentBucket() throws Exception {
+    // This test is incorrect because the fake-bucket in the url does not exist
+    mockDetectSafeSearch.detectSafeSearchGcs("gs://fake-bucket");
+  }
+
+  @Test
+  public void testDetectSafeSearchGcs_noFileWithPath() throws Exception {
+    // This test is incorrect because the file missing-image.jpg does not exist in the keyframe-images bucket
+    mockDetectSafeSearch.detectSafeSearchGcs("gs://video-vigilance-videos/missing-image.jpg");
```
vkkpng-g, line 99: I was wondering how calling these methods are throwing exceptions. It seems like the class u defined above MockDetectSafeSearchGcs does not throw any errror?

capstone/src/test/java/com/google/sps/DetectSafeSearchGcsTest.java
cassmarcussen: Removed

capstone/src/test/java/com/google/sps/DetectSafeSearchGcsTest.java
cassmarcussen: Sorry, I should have deleted all the "throws Exception" in the code. I just deleted them. None of the methods will throw exception. "throws Exception" just tells the user that an exception may be thrown, but now no exceptions will be thrown, so they can be removed.

capstone/src/test/java/com/google/sps/DetectSafeSearchGcsTest.java
```diff
@@ -0,0 +1,106 @@
...
+    mockSafeSearchResults.put("violence", "UNLIKELY");
+    mockSafeSearchResults.put("racy", "VERY_UNLIKELY");
+
+    // Specify which functions of mockDetectSafeSearchGcs
+    when(mockDetectSafeSearch.detectSafeSearchGcs(anyString())).thenReturn(mockSafeSearchResults);
+
+  }
+
+  @Test
+  public void testDetectSafeSearchGcs_incorrectBucketPathOneSlash() {
```
wdi-gg, line 69: Current tests won't work, since even if something is broken in DetectSafeSearchGcs.java, these tests would still pass, right?
The best way to do this will be mocking client.batchAnnotateImages() and make it return BatchAnnotateImagesResponse that you defined, and then test if your logic handles those responses correctly.
If that's too hard to figure out, you can also extract the API calling logic into a wrapper method and mock your wrapper method (as Vikki mentioned in chat).

cassmarcussen: That makes sense. In my understanding of what methods to mock, I was mocking the wrong method (detectSafeSearchGcs). I will work on making this change.

cassmarcussen:
Sorry, I was trying to fix merge conflicts between this and cem-image-serving-and-effect and accidentally merged. For Di and Vikki, please still review this code and I will make the changes accordingly. Sorry about this!

PR #17, opened 2020-07-09
Upload video

sharond106:
This PR uploads a video file to a google cloud bucket. The HTML form is submitted with ajax (on the javascript file) to prevent the page from refreshing. With a normal POST request, the redirect response causes the entire page to refresh, and the input file would disappear. 

GET: retrieves the Cloud bucket url containing the video file (or the most recently uploaded video file if there are multiple)
POST: uploads the video file to Datastore and Cloud bucket

The GET and POST method's code is in another class, VideoUpload.java, so I can easily test them later.

Process:
User uploads a file to the webpage -->  User hits submit button to submit the form --> ajax code on the javascript file overrides the normal POST request and submits the form itself --> java servlet gets url from Blobstore and stores it in Datastore --> returns to ajax code in javascript file (doesn't need to do anything with the response)

The ShotsServlet.java's GET method is now able to use a real url passed into the request. This url will be used by the Video Intelligence API to detect shots from.

PR #20, opened 2020-07-13
Store shot times along with images

sharond106:
This PR makes the small change of storing extracted images along with their shot times, for displaying with their final effects


capstone/src/main/webapp/js/extract-images-script.js
```diff
@@ -183,7 +191,12 @@ function displayFrame(img, secs, event) {
   //    
   //    The next frame would be at 15 s., but since this is > 12 s., do not capture another frame
   if (frameInterval != -1 && (secs + frameInterval <= video.duration)) {
-    captureFrame(video.src, secs + frameInterval);
+    const shotObject = {
+      start: secs, 
+      middle: Math.round(secs + frameInterval),
+      end: Math.round(secs + frameInterval)
+    };
```
sharond106, line 198: What should the shot start/end time be for when I'm capturing image frames based on a user inputted time interval? Currently they're like (interval = 5s):
Shot 1: 0 - 5s
Shot 2: 5s - 10s
Shot 3: 10s - 15 s

capstone/src/main/webapp/js/extract-images-script.js
```diff
@@ -14,7 +14,7 @@
 
 /** Javascript functions for extracting images from video */
 
-// Array of times to keyframe images at
+// Array of shot objects to keyframe images at
```
cassmarcussen, line 17: Unfinished comment? (should there be something after the 'at'?)

capstone/src/main/webapp/js/extract-images-script.js
```diff
@@ -42,14 +42,19 @@ function getShots() {
...
 
-    // Display each shot's times in a list and add the middle time of each shot to keyTimes array
+    // Display each shot's times in a list and add start, middle, end time of each shot to keyTimes array
     for (const shot of shots) {
       const listElement = document.createElement("li");
       const textElement = document.createElement("span");
       textElement.innerHTML = "<b>Shot " + count + ": <b>" + shot.start_time + " - " + shot.end_time;
       listElement.appendChild(textElement);
       list.append(listElement);
-      keyTimes.push((shot.start_time + shot.end_time) / 2.0);
```
cassmarcussen, line 52: Use camelcase, i.e. startTime, endTime, since in JS camelcase is the standard and for consistency

capstone/src/main/webapp/js/extract-images-script.js
cassmarcussen: Maybe we can change the html on the results page to say "interval start" and "interval end" if the user specifies this option? Is there any way we could pass information around about which option the user chooses?

sharond106: Yes i think we can just set a javascript variable to do that!

capstone/src/main/webapp/js/extract-images-script.js
sharond106: Oh I did this again.. I'll change it

capstone/src/main/webapp/js/extract-images-script.js
sharond106: fixed

capstone/src/main/webapp/js/extract-images-script.js
sharond106: I'm actually already doing that with my frameInterval variable. It's set to -1 when the interval method is not being used and will be a positive number when it is

sharond106:
@wdi-gg PR #17 is based off an earlier branch that also doesn't have these changes right now, but once I merge everything they will all have it!

PR #28, opened 2020-07-23
Optimize Java servlet calls for the Upload Servlet and DetectSafeSearchGcs

cassmarcussen:
### Changes Made
Moved the retrieval of the SafeSearch effect of each keyframe image to the Upload Servlet's GET method. This has 2 benefits: 
- First, it moves logic to the server-side Java functions, leaving the JavaScript code to just have the role of rendering content.
- Second, it reduces the workload required to display the effect, since we aren't passing an array of Keyframe Images between Java servlets through JavaScript calls. It is therefore cleaner in design and more efficient.


capstone/src/main/java/com/google/sps/data/KeyframeImage.java
```diff
@@ -10,12 +12,17 @@
   private String cloudBucketUrl;
   // The following three int variables are times in number of seconds.
```
rwackes, line 13: Update comment since there are no three int variables

vkkpng-g: +1 It seems ur comment is still  "// The following three (change to 2)  int variables are times in number of seconds."

capstone/src/main/webapp/js/image-effect-results.js
```diff
@@ -370,15 +369,6 @@ async function fetchBlobstoreKeyframeImages(shouldDisplayOnlyFlaggedImages) {
 
       var arrayOfKeyframeImages = JSON.parse(keyframeImages);
```
vkkpng-g, line 370: Nit: seems like u can remove 2 spaces to align with 238

cassmarcussen: This is indented 2 spaces because it is in a 'then' (line 368). For the PR GitHub collapsed that code, but it is indented by 2 spaces from there.

PR #30, opened 2020-07-29
Update UI for Audio Effect Results and Main Page

rwackes:
In this PR, I made the following changes: 
- Display label, meter, and score in the same line (Screenshot: https://screenshot.googleplex.com/gn17RVNF4WH)
- Transform meter rating to: Very Unlikely, Unlikely, Possible, Likely, Very Likely
- Display original numerical scores when user hovers over the icon next to the rating
- Give user the option to see the computer-generated transcription ([collapsible](https://screenshot.googleplex.com/SviM0uuhv2Q))
- Fix the position of tooltip text
- Fix the profile images on /index.html so they do not appear stretched (Screenshot: https://screenshot.googleplex.com/2XNZet5etA2)

Screenshot of UI: https://screenshot.googleplex.com/NUJ6Pgr6xfA


capstone/src/main/webapp/js/extract-audio-script.js
```diff
@@ -99,6 +91,156 @@ function createAudioTranscription() {
...
+}
+
+/**
+ * Create html for a button that will expand a collapsible div containing the transcript
+ * of the user's audio and the confidence level.  
+ */
+function createCollapsibleTranscript(effectObj) {
+  var button = '<label class="conf-label">Our confidence level in this transcription is ' + effectObj.confidence + '%.' 
+      + '<div class="tooltip-info"> '
+        + '<i class="fa fa-info-circle" aria-hidden="true"></i> '
```
sharond106, line 233: For this function and createTranscriptHTML() could you put the HTML code on the results.html page and just set the css display: none initially? Then you can insert effectObj.transcription into it and set display: inline-block/ block when you want to show it instead of printing all the HTML code from javascript.

capstone/src/main/webapp/js/extract-audio-script.js
```diff
@@ -99,6 +91,156 @@ function createAudioTranscription() {
...
+ * by Perspective API
+ * @param effectObj: the response from the servlet
+ */
+function createScoresHTML(effectObj) {
+  var content = '<h2 class="card-title">Effect of the audio</h2>'
+    + '<div class="card-text" id="card-image">'
+    + '<p>Attributes are scored from 0 - 10, with 0 being most unlikely to be perceived as the attribute and 10 being most '
+    + 'likely to be perceived as the attribute. Scores below 2 are classified as Very Unlikely, between 2 and 4 are Unlikely, between 4 and 6 '
+    + 'are Possible, between 6 and 8 are Likely, and above 8 are Very Likely. Values greater than or equal to 6 are flagged.</p>'
+    + createMeterHTML("toxicity", "Toxicity", effectObj.TOXICITY)
```
cassmarcussen, line 105: This was suggested to me in one of my PRs (for my create html function), but maybe we can move each of the createMeterHTML calls to above Line 100, defining them with vars.

capstone/src/main/webapp/js/extract-audio-script.js
```diff
@@ -99,6 +91,156 @@ function createAudioTranscription() {
...
+}
+
+/**
+ * Return the definition of a given Perspective API attribute, as outlined by the API developers
+ * Reference: https://github.com/conversationai/perspectiveapi/blob/master/2-api/models.md#all-attribute-types
+ * @param attributeId: a string to identify the attribute whose definition we seek
+ */
+function getDefinition(attributeId) {
+  var definition = '';
+  if (attributeId.localeCompare("confidence") == 0) {
```
cassmarcussen, line 146: Is there any way to break up this change of if / else statements? Maybe with a switch statement?

capstone/src/main/webapp/js/extract-audio-script.js
```diff
@@ -99,6 +91,156 @@ function createAudioTranscription() {
...
+  if (score < 2) {
+    value = 1;
+  } else if (score < 4) {
+    value = 2;
+  } else if (score < 6) {
+    value = 3;
+  } else if (score < 8) {
+    value = 4;
+  } else {
+    value = 5;
```
cassmarcussen, line 183: Can we replace this with a switch statement or an enum?

capstone/src/main/webapp/js/extract-audio-script.js
```diff
@@ -99,6 +91,156 @@ function createAudioTranscription() {
...
+
+/**
+ * Returns the likelihood that the audio will be perceived as an attribute based on 
+ * the numerical score returned by Perspective API.
+ * Attribute summary scores below 2 are very unlikely, below 4 are unlikely, below 6 
+ * are possible, below 8 are likely, and between 8 and 10 are very likely. 
+ */
+function getScoresAsLikelihood(score) {
+  var likelihood = '';
+  if (score < 2) {
```
cassmarcussen, line 196: Same here with replacing with switch statement or enum

capstone/src/main/webapp/css/results.css
```diff
@@ -0,0 +1,244 @@
+.results {
+  margin-left: auto/* 15% */;
+  margin-right: auto;
+  padding: 80px 0px;
+  width: 1000px;
+}
+
+img {
+  width: 700px;
```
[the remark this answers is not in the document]
rwackes: Noted. When merging and solving merge conflicts, will keep yours.

capstone/src/main/webapp/js/extract-audio-script.js
rwackes: Done!

capstone/src/main/webapp/js/extract-audio-script.js
rwackes: Switched to switch statements. Done.

capstone/src/main/webapp/js/extract-audio-script.js
rwackes: Switch statements don't support ranges that well so I used if/else.

capstone/src/main/webapp/js/extract-audio-script.js
rwackes: Same here. Switch statements don't support ranges that well so I used if/else.

capstone/src/main/webapp/js/extract-audio-script.js
wdi-gg: Can we just do (score / 2) and map it to 0-4? +1 if meter requires 1-5. May need to handle score==10 if it's a possible value?

capstone/src/main/webapp/js/extract-audio-script.js
```diff
@@ -99,6 +92,167 @@ function createAudioTranscription() {
...
+      + '<div class="tooltip-orig-score"> '
+        + '<i class="fa fa-info-circle" aria-hidden="true"></i> '
+        + '<span id="tooltiptext-orig-score">Your video\'s toxicity score was a ' + score + ' out of 10. Meaning, the likelihood '
+        + 'that your video will be perceived as toxic by your audience is ' + (score*10).toFixed(1) + '%.</span>'
+      + '</div>'
+    + '</label> ';
+  return meterContent;
+}
+
+/**
```
wdi-gg, line 141: Let's try using dictionary rather than function? https://google.github.io/styleguide/jsguide.html#jsdoc-enum-and-typedef-comments

capstone/src/main/webapp/js/extract-audio-script.js
wdi-gg: Same here haha. If score can be 0-4, just create a global array and use score as index, rather than a function.

capstone/src/main/webapp/js/extract-audio-script.js
rwackes: Switched to a switch statement using Math.floor(score / 2) + 1 and a case to handle when score == 10. Could not use enum here because it is not valid syntax
const Value =  {
  0: 1,
  1: 1,
  2: 2,
  ...
}; 
is not supported.

Done!

rwackes: Changed if/else to switch statement checking Math.floor((score/2) + 1) and with a case to handle when score == 10. Used switch over enum since enum does not support the following syntax.
const ScoreAsValue = {
  0: 1,
  1: 1,
  2: 2,
  etc...
};

Switch made more sense.

Done.

capstone/src/main/webapp/js/extract-audio-script.js
rwackes: Changed if/else to switch statement checking Math.floor((score/2) + 1) and with a case to handle when score == 10. Used switch over enum since enum does not support the following syntax.
const ScoreAsLikelihood = {
  0: 'Very Unlikely',
  1: 'Very Unlikely',
  2: 'Unlikely',
  etc...
};

Switch made more sense.

Done.

capstone/src/main/webapp/js/extract-audio-script.js
rwackes: Replaced getDefinition() with enum. Done.

capstone/src/main/webapp/js/extract-audio-script.js
```diff
@@ -99,11 +89,178 @@ function createAudioTranscription() {
...
+ * Returns a value properly formatted for the new meter range corresponding to the
+ * attributes' summary score returned by Perspective API.
+ * Attribute summary scores below 2 are formatted to 1, below 4 are formatted to 2, below 6 
+ * are formatted to 3, below 8 are formatted to 4, and between 8 and 10 are formatted to 5. 
+ */
+function getScoreAsValue(score) {
+  var value = '';
+  var range = Math.floor((score/2) + 1);
+  switch(range) {
```
wdi-gg, line 198: It's easy to handle 1-5, what about 1-100? Can we just use "toString()"?

capstone/src/main/webapp/js/extract-audio-script.js
```diff
@@ -99,11 +89,178 @@ function createAudioTranscription() {
...
+  return value;
+}
+
+/**
+ * Returns the likelihood that the audio will be perceived as an attribute based on 
+ * the numerical score returned by Perspective API.
+ * Attribute summary scores below 2 are very unlikely, below 4 are unlikely, below 6 
+ * are possible, below 8 are likely, and between 8 and 10 are very likely. 
+ */
+function getScoresAsLikelihood(score) {
```
wdi-gg, line 230: As mentioned in the previous review: create a global array and map score to index.
e.g. 
var likelihood = ["Very Unlikely", "Unlikely"...];
var index = Math.floor(score/2);
// handle corner case, etc.
return likelihood[index];

rwackes: Created two global arrays and mapped the numerical score returned from the servlet (a number from [0, 10]) to an index value from [0, 6] using Math.floor((score/2)+1). Then, I accessed the global arrays for the corresponding likelihood value and meter value using the index [0, 6]. Done!

capstone/src/main/webapp/js/extract-audio-script.js
rwackes: Replaced these switch statements with global array so no need to use toString() anymore. Done.

capstone/src/main/webapp/js/extract-audio-script.js
wdi-gg: When will "Unknown" be returned? I think we need to check validity of the index, same for getScoreAsValue().
e.g.
if index valid, return scoreAsLikelihood[index];
else return "Unknown";

rwackes: Hmmm yeah the way I was calculating it, it would have never accessed index 0 for "Unknown" or the value 0. These "invalid" cases would have never been used by my application in any edge case. Thanks for pointing this out! 
I modified the way I calculate index and the global arrays so that each index is now a valid real value that may be used by our application. Done in new commit. 

Note: I don't need to check "validity" per se of the index in my code because there is no chance of the score returned by Perspective API/my servlet being outside the range of [0, 10] aka being "invalid". Perspective ONLY returns scores from [0, 1] and I multiply that by ten so my servlet ONLY returns scores from [0, 10]. I do not use the "Unknown" scoring metric like Cassandra. The values her API returns includes an "Unknown". The values my API returns are numerical and do not map to "Unknown" in any case. If Perspective API experiences an error, it throws an error, so I would display an error message, whereas if Cassandra's API experiences an error in scoring an image on one of the attributes, it lists as "Unknown." Did I explain well?

wdi-gg: SG!

PR #31, opened 2020-07-30
Post captured images

sharond106:
In this PR, I send a POST request for each image to a servlet that uploads it to Datastore and a Cloud bucket.

More specifically, I added a hidden form on upload.jsp and an ajax request in extract-images-script.js that submits the hidden form, which posts a captured image to the Blobstore url created on upload.jsp

[This code was already in the testing-merged-content branch, but I had to copy it into this one to merge with master]

### DIFF
--- a/capstone/src/main/webapp/js/extract-images-script.js
+++ b/capstone/src/main/webapp/js/extract-images-script.js
@@ -31,6 +31,10 @@ var frameNum = 0;
 var submitting = false;
 var submitted = false;
 
+// Variables for submitting the image
+var blob;
+var blobShotObject;
+
 // Ajax code that submits video file form
 $(document).ready(function() {
   // When the user submits the form to upload a video, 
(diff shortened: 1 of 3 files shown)


capstone/src/main/webapp/upload.jsp
```diff
@@ -94,6 +100,19 @@
       <br><br>
       <div id="transparent-placeholder"></div>
       <br>
+      <!-- This form is hidden to the user (Images must be submitted from jsp file) -->
+      <form id="post-keyframe-img" method="POST" enctype="multipart/form-data" action="<%= imageUploadUrl %>">
+        <p>Timestamp:</p>
+        <textarea id="timestamp" name="timestamp"></textarea>
+        <p>Start Time:</p>
```
cassmarcussen, line 107: For the form, should we remove startTime and endTime and add in isManuallySelected?

sharond106: oh right I forgot about that! I haven't been able to test without our code merged, but I will also add a manuallyCaptured field.

capstone/src/main/webapp/js/extract-images-script.js
```diff
@@ -270,11 +274,38 @@ function postFrame(canvas, shot) {
   canvas.toBlob(function(thisblob) {
     img.src = URL.createObjectURL(thisblob);
```
vkkpng-g, line 275: move 2 spaces left

PR #33, opened 2020-07-30
Merge VideoUploadServlet and ShotsServlet 

sharond106:
Instead of fetching the url containing the video from VideoUploadServlet, I fetch it and get the shots returned from the API together from the ShotsServlet, turning 2 fetch calls into 1. I'm also now using the ShotsServlet to POST the video, since it did not have a POST method before. This allowed me to delete the VideoUploadServlet entirely.

To do this, I had to change the return value of getUrl() in VideoUpload.java to return a Map instead of a String formatted as a json so that the GET method in ShotsServlet could easily parse it. I had to modify my VideoUpload tests as well to receive a Map instead of a String


capstone/src/main/webapp/js/extract-images-script.js
```diff
@@ -107,27 +107,19 @@ function getShots() {
...
-  fetch("/video-upload?name=" + datastoreName).then(response => response.json()).then(jsonObj => {
-    // If there was an error getting the url, use backup shots option
-    if (jsonObj.error) {
-      checkForShots();
-    } else {
-      // Send the bucket url to the Video Intelligence API and get shot times
-      fetch("/shots?url=gs:/" + jsonObj.url).then(response => response.json()).then(shots => {
-        // Remove loading message
-        message.innerHTML = "";
+  fetch("/shots?name=" + datastoreName).then(response => response.json()).then(shots => {
```
cassmarcussen, line 110: Split up the chain of fetch/then/then to multiple lines.

capstone/src/main/webapp/js/extract-images-script.js
```diff
@@ -107,27 +107,19 @@ function getShots() {
...
-            manuallyCaptured: false
-          };
-          keyTimes.push(shotObject);
-        }
-      }).then(() => checkForShots());
+    // Add the middle time of each shot to keyTimes array
+    for (const shot of shots) {
+      const shotObject = {
+        timestamp: Math.round((shot.startTime + shot.endTime) / 2.0),
+        manuallyCaptured: false
```
cassmarcussen, line 118: Can you change to isManuallyCaptured, for naming consistency with DataStore components?

sharond106: i'm going to have to do this after merging all my branches since this will cause changes in other branches as well

capstone/src/main/webapp/js/extract-images-script.js
sharond106: fixed!

capstone/src/main/java/com/google/sps/servlets/ShotsServlet.java
```diff
@@ -24,31 +33,40 @@
...
 import javax.servlet.http.HttpServlet;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 import java.util.ArrayList; 
+import java.util.List;
+import java.util.Map;
+import java.util.HashMap;
 
 /** Servlet that gets shot changes for uploaded video from Video Intelligence API*/
 @WebServlet("/shots")
```
vkkpng-g, line 49: Did we update the test for ShotsServlet?

sharond106: No I don't think I need to. The VideoUploadTest was testing the old servlet whose code basically just got moved to ShotsServlet. ShotsServlet now makes the calls to VideoUpload instead of the old servlet.

PR #34, opened 2020-07-30
Refactor JS and Avoid Timeout Error

rwackes:
In this PR, I did the following:
- changed the value that we use to "flag" the audio from 5 to 6 to match the new attribute rating of Very Unlikely (scores between 0 and 2), Unlikely (scores between 2 and 4), Possible (scores between 4 and 6), Likely (scores between 6 and 8), and Very Likely (scores between 8 and 10)
- changed our application's module from automatic scaling (where requests time out at 1 minute) to manual scaling (where requests time out at 10 minutes) so that our application can process longer videos without running into the DeadlineExceededException (aka the timeout error) unless it takes over 10 minutes
- moved JS code into helper functions to improve readability and reduce each function size
- combine two fetch statements into one so that instead of fetching the url (1) and then fetching the audio effect (2), I fetch the url and the audio effect in the same servlet. This change is to correspond with Sharon's code found in https://github.com/cassmarcussen/video-vigilance/pull/33


capstone/src/main/java/com/google/sps/servlets/AudioEffectServlet.java
```diff
@@ -40,31 +43,31 @@
...
-    } catch (DeadlineExceededException e) {
-      // GAE abruptly broke out of the try block because the request timed out (took longer than 60 seconds).
-      results.put("error", "timeout");
+    // Get GCS bucket url that has user-submitted video.
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    VideoUpload videoUpload = new VideoUpload();
+    Map<String, String> urlErrorMap = videoUpload.getUrl(datastore, request.getParameter("name"));
+    // If there's no error fetching the url, perform operations to get audio's effect.
+    if (urlErrorMap.get("error").isEmpty()) {
+      String gcsUri = "gs:/" + urlErrorMap.get("url");
```
sharond106: i'm not sure if isEmpty() works the same as just checking if the "error" value is the empty string "", but I do manually put in an empty string for the error key, if you need to change this

sharond106 (approved):
So much cleaner! I just added a comment in your servlet but everything looks good! Also congrats on figuring out the silly timeout error

capstone/src/main/java/com/google/sps/servlets/AudioEffectServlet.java
cassmarcussen: nit: make "gs:/" a constant String, and call this constant String in Line 54

cassmarcussen (approved):
LGTM!

capstone/src/main/java/com/google/sps/servlets/AudioEffectServlet.java
rwackes: Should not be a problem. Works the same.

rwackes: Done!

wdi-gg (approved):
Thanks for figuring out the timeout issue! Good job!

vkkpng-g (approved):
LGTM good job figuring out the timeout!

And really like the cleanup you did.

